PR #9, opened 2018-11-27
Add controller for syncing autoscaling groups

### LINKED ISSUE #3: Implement node count reconciliation
See https://github.com/containership/internal-rfc/pull/3 for more details.

This particular ticket is only for the informers + reconciliation loop(s) needed to reconcile node count to be within the desired min/max bounds of the AutoScalingGroup (ASG) CustomResource (CR).

There should be no metrics integration for this ticket.

ashleyschuett:
### Description

This PR adds the functionality to register and initialize a simple "dump" autoscaling engine. It also creates a controller to queue an autoscaling group resource on add or edit, as well as enqueue the autoscaling group associated with a node on node add or delete. 

This controller only works on > 1.11.x 

* Fixes #3 

 ### Testing

Ran the controller with an ASG that had a min greater than the node count for the cluster and it triggered the registered containership engine. 

I also ran the controller with the number of nodes in the bounds of min and max and it nooped. 

Created tests for some of the controllers simpler functions which can be ran using `make test`

 #### Setup

`KUBECONFIG=$HOME/.kube/config go run cmd/autoscalingengine/autoscalingengine.go`

 #### Instructions


cmd/autoscalingengine/app/autoscalingengine.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/containership/cerebral/pkg/buildinfo"
+	cerebral "github.com/containership/cerebral/pkg/client/clientset/versioned"
+	cinformers "github.com/containership/cerebral/pkg/client/informers/externalversions"
+	"github.com/containership/cerebral/pkg/controller"
+
+	"github.com/containership/cluster-manager/pkg/log"
+)
+
+// NewAutoScalingEngineCommand creates a command to run the autoscaling engine
+func NewAutoScalingEngineCommand() *cobra.Command {
```
ashleyschuett: This might be overkill and totally not needed.. curious what other peoples thoughts are on this implementation.

pkg/autoscalingengine/autoscalingengine.go
```diff
@@ -0,0 +1,103 @@
...
+	autoscalingEngines     = make(map[string]Interface)
+)
+
+// Interface species the functions that an autoscaling group must implement
+type Interface interface {
+	SetTargetNodeCount(nodeSelectorList labels.Selector, numNodes int, heuristic string) error
+}
+
+// RegisterAutoscalingEngine initializes an instances of the autoscaling engine
+func RegisterAutoscalingEngine(name, config string, autoscalingengine Factory) {
```
ashleyschuett: This creates initializes the engine when it is registered. This way if cerebral registers more than one engine each autoscaling group can then specify and use the engine of its choice.

pkg/autoscalingengine/autoscalingengine_test.go
```diff
@@ -0,0 +1,6 @@
+package autoscalingengine
+
+// TODO tests following functions
```
ashleyschuett: If we like this approach I'll write tests for the functions but didn't want to take the time without some initial feedback on the design.

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,330 @@
...
+	return asg.updateAutoScalingGroupStatus(autoScalingGroup)
+}
+
+func (asg *AutoScalingGroupThresholdSyncController) updateAutoScalingGroupStatus(autoScalingGroup *cerebralV1alpha1.AutoScalingGroup) error {
+	// NEVER modify objects from the store. It's a read-only, local cache.
+	// You can use DeepCopy() to make a deep copy of original object and modify this copy
+	// Or create a copy manually for better performance
+	asgCopy := autoScalingGroup.DeepCopy()
+	asgCopy.Status.LastUpdatedAt = time.Now().Unix()
+	// .UpdateStatus can only be used in 1.11 and greater
```
ashleyschuett: Are we going to support 1.10.x? Or is it safe to use `.UpdateStatus` since we will only be supporting 1.11 and greater?

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,330 @@
...
+		l, _ := labels.NewRequirement(key, selection.Equals, []string{value})
+		selector = selector.Add(*l)
+	}
+
+	return selector
+}
+
+// findNodesASG goes through each autoscaling group and checks to see if the
+// nodeSelector matches the node labels passed into the function
+// TODO: what if the user messes up and there's more than one ASG that matches,
```
ashleyschuett: Thoughts?

pkg/autoscalingengine/autoscalingengine.go
ashleyschuett: Is this something we would want to make into a grpc call instead of having people register on init and recreating the main command to create and use their own engines?

cmd/autoscalingengine/app/autoscalingengine.go
mattkelly: I personally don't see a purpose for it since we're not consuming any flags or config files

cmd/autoscalingengine/app/autoscalingengine.go
```diff
@@ -0,0 +1,111 @@
...
+
+	cmd := &cobra.Command{
+		Use:  "auto-scaling-manager",
```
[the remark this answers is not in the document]
ashleyschuett: I'm in favor of that oo, but then I feel like we need to change the CRD kind to be `AutoscalingEngine` ?

pkg/autoscalingengine/containership/containership.go
```diff
@@ -0,0 +1,32 @@
...
+
+// NewAutoscalingEngine creates a new instances of the containership autoScaling
+// engine
+func NewAutoscalingEngine() (autoscalingengine.Interface, error) {
+	return &containershipAutoScalingEngine{}, nil
+}
+
+func init() {
+	// TODO: should this be a flag, or an env var?
+	pathToConfig := ""
```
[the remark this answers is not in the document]
ashleyschuett: I was kind of thinking just for auth information. then users could load the file into a struct to use instead of specifying env var for endpoints, routes, tokens, etc.

pkg/controller/autoscalinggroup.go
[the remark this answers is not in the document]
ashleyschuett: This is also why I was thinking Ownerref might be useful, go through each ASG once for each node max and assign it. then when a node is queued just make sure the ASG owner ref still matches the node labels since labels can change which would be less. 

But that also doesn't answer the question of multiple ASGs matching

cmd/autoscalingengine/app/autoscalingengine.go
```diff
@@ -0,0 +1,111 @@
...
+		kubeclientset, kubeInformerFactory, cerebralclientset, cerebralInformerFactory)
+
+	kubeInformerFactory.Start(stopCh)
+	cerebralInformerFactory.Start(stopCh)
+
+	if err = autoscalingGroupController.Run(1, stopCh); err != nil {
+		log.Fatalf("Error running controller: %s", err.Error())
+	}
+
+	panic("unreachable")
```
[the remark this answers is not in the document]
ashleyschuett: I put this here since the function specifies a return value of type error. It feels weird to return nil, or an error here though.. :/

pkg/autoscalingengine/autoscalingengine.go
ashleyschuett: @mattkelly any thoughts on this?

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,330 @@
...
+
+	asgLister clisters.AutoScalingGroupLister
+	asgSynced cache.InformerSynced
+
+	workqueue workqueue.RateLimitingInterface
+}
+
+// NewAutoScalingGroupThresholdSyncController returns a new controller to watch
+// that the nodes associated with autoscaling groups stay between the min and max specifed
+func NewAutoScalingGroupThresholdSyncController(kubeclientset kubernetes.Interface,
```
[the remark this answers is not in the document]
ashleyschuett: Im not sure what you mean by this?

pkg/controller/autoscalinggroup.go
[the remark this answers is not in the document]
ashleyschuett: @normanjoyner any thoughts?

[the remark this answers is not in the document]
mattkelly: FWIW I should probably clarify that iterating over the entire range (which will be small) every time is obviously of negligible performance impact, just a bit weird.

I think I'd prefer to just do that and not do anything with `ownerRef`s. It feels weird to me to essentially adopt resources (nodes) that we, in most cases, are not responsible for creating.

pkg/autoscalingengine/containership/containership.go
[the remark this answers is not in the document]
mattkelly: Wouldn't a ConfigMap or Secret be more appropriate for that, versus requiring users to place a config file in the Docker image?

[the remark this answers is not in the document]
ashleyschuett: Thats what I was referring to. If they create a configmap, or secret and then they could mount the whole config map or secret using kubernetes volumes, and specify the path to where it is mounted. Or are we still referring to something different?

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,365 @@
...
+		return errors.Wrapf(err, "error syncing autoscaling group %q (has been requeued %d times)", key, asg.workqueue.NumRequeues(key))
+	}
+
+	asg.workqueue.Forget(key)
+	log.Infof("Dropping autoscaling group %q out of the queue: %v", key, err)
+	return err
+}
+
+// enqueueNode enqueues the ASG that matches the enqueued node's labels
+func (asg *AutoscalingGroupController) enqueueNode(obj interface{}) {
```
mattkelly: This should be named differently since it does not enqueue Nodes. Maybe `enqueueASGForNode()`?

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,365 @@
...
+		return nil
+	}
+
+	if isScaleUpEvent(numNodes, desired) {
+		asg.recorder.Event(autoScalingGroup, corev1.EventTypeNormal, events.ScaledUp, fmt.Sprintf("Autoscaling group %s node count (%d) was not within min and max bounds and was scaled up to %d", autoScalingGroup.Name, numNodes, desired))
+	} else {
+		asg.recorder.Event(autoScalingGroup, corev1.EventTypeNormal, events.ScaledDown, fmt.Sprintf("Autoscaling group %s node count (%d) was not within min and max bounds and was scaled down to %d", autoScalingGroup.Name, numNodes, desired))
+	}
+
+	return asg.updateAutoScalingGroupStatus(autoScalingGroup)
```
ashleyschuett: should this actually return on error? Seems weird to make another call to the ase, but also seems wrong to be okay with not updating the last update status since thats used for the cooldown period

pkg/apis/cerebral.containership.io/register.go
```diff
@@ -2,5 +2,5 @@ package metricsbackendcerebralcontainershipio
 
 const (
 	// GroupName is the group name for this API group
-	GroupName = "metricsbackend.cerebral.containership.io"
+	GroupName = "cerebral.containership.io"
```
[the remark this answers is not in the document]
mattkelly: The package name should be fixed as well - can't comment because not close enough to diff

deploy/crds/cerebral-autoscaling-group.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
mattkelly: On master, this dir is called `deploy/crd`. Can you place this file in the same dir?

Gopkg.toml
```diff
@@ -12,8 +12,39 @@ required = [
...
+[[constraint]]
+  name = "github.com/containership/cluster-manager"
+  version = "v4.0.1"
+
+[[constraint]]
+  name = "sigs.k8s.io/controller-runtime"
+  version = "v0.1.7"
+
+[[override]]
+  branch = "master"
```
mattkelly: Why are we pulling in master of a dependency / why is this override needed? Can we pin to a commit and add a comment?

Gopkg.toml
```diff
@@ -12,8 +12,39 @@ required = [
...
+  name = "github.com/golang/groupcache"
+
```
mattkelly: Extra spaces here

pkg/apis/cerebral.containership.io/v1alpha1/types.go
```diff
@@ -31,3 +31,42 @@ type MetricsBackendList struct {
...
+	Engine         string            `json:"engine"`
+	CooldownPeriod int               `json:"cooldownPeriod"`
+	Suspend        bool              `json:"suspend"`
+	MinNodes       int               `json:"minNodes"`
+	MaxNodes       int               `json:"maxNodes"`
+}
+
+// AutoScalingGroupStatus is the status for a autoscaling group
+type AutoScalingGroupStatus struct {
+	LastUpdatedAt int64 `json:"LastUpdatedAt"`
```
mattkelly: JSON key shouldn't start with uppercase

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,370 @@
...
+	//
+	delayBetweenRequeues = 30 * time.Second
+
+	// Requeue autoscaling groups if there is an error up to 5 times,
+	maxRequeues = 5
+)
+
+// AutoscalingGroupController is a controller for scaling a autoscaling
+// group based on min and max
+type AutoscalingGroupController struct {
```
mattkelly: Does just `AutoscalingController` make more sense?

ashleyschuett: the only reason I don't like that is because I think its less clear since we are also adding a "controller" (event though it's not really) for making changes to autoscaling via metrics.

pkg/events/events.go
```diff
@@ -0,0 +1,8 @@
+package events
+
+const (
+	// ScaledUp event is created when an AutoScalingGroup is scaled up
```
mattkelly: `Autoscaling`, and below

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,379 @@
...
+	if a == nil {
+		return
+	}
+
+	agc.enqueueAutoscalingGroup(a)
+}
+
+// enqueueAutoscalingGroup enqueues an autoscalinggroup object.
+func (agc *AutoscalingGroupController) enqueueAutoscalingGroup(obj interface{}) {
+	log.Info("added to queue")
```
mattkelly: `Debug` or just remove. I'm in favor of removing because of line 253 which is more accurate.

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,379 @@
...
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		log.Error("Error enqueueing autoscaling group", err)
+		return
+	}
+	log.Info("added to workqueue ", key)
+	agc.workqueue.AddRateLimited(key)
+}
+
+// getDesiredNodeCount returns curr if curr is between the min and max bounds,
+// if its below the bounds it will return min, if its above the bounds it will
```
mattkelly: `it's`

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,379 @@
...
+	return curr
+}
+
+func isScaleUpEvent(curr, desired int) bool {
+	return curr < desired
+}
+
+// syncHandler surveys the state of the systems, checking to see that the current
+// number of nodes selected by an autoscaling groups node selector is within the
+// autoscaling groups bounds for min and max
```
mattkelly: And taking actions to reconcile if not

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,379 @@
...
+	if numNodes == desired {
+		return nil
+	}
+
+	engine, err := agc.autoscalingengine.GetAutoscalingEngine(engineName)
+	if err != nil {
+		return err
+	}
+
+	scaled, err := engine.SetTargetNodeCount(ns, desired, "")
```
mattkelly: The heuristic passed should depend on whether it's a scale up or scale down. Along those lines, the heuristic would normally come from an AutoscalingPolicy, but this scale event isn't from a metric alert and thus can't be associated with a specific Policy. What's the expected behavior in this case? 🤔

pkg/controller/autoscalinggroup.go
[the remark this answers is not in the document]
mattkelly: Let's resolve this before merge

pkg/controller/autoscalinggroup_test.go
```diff
@@ -0,0 +1,178 @@
...
+	name     string
+	curr     int
+	min      int
+	max      int
+	expected int
+}
+
+var desiredCountTests = []*buildDesiredCountTest{
+	&buildDesiredCountTest{
+		name:     "Current node count is with in min and max bounds, return current count",
```
mattkelly: `within`

pkg/controller/autoscalinggroup_test.go
```diff
@@ -0,0 +1,178 @@
...
+	},
+	&buildNodesLabelSelector{
+		name: "check multiple labels",
+		labels: map[string]string{
+			"key-1": "value-1",
+			"key-2": "value-2",
+		},
+		expectedString: "key-1=value-1,key-2=value-2",
+	},
+	// TODO: are we supporting empty?
```
mattkelly: Good question. Is it possible to block the nodeSelector field being empty via OpenAPI validation?

cmd/autoscalingengine/autoscalingengine.go
```diff
@@ -0,0 +1,57 @@
+package main
```
mattkelly: I don't think this file structure is correct. Shouldn't this be more like `cmd/cerebral/cerebral.go`, since this is the main (and today the only) binary?

pkg/controller/autoscalinggroup_test.go
ashleyschuett: I believe so, but when I was thinking about it I wasn't sure we would want to block it. If no selector matches everything maybe the user just wants to make that all there nodes within a cluster stays within a range of min and max, so it might be a valid use case?

pkg/apis/cerebral.containership.io/v1alpha1/types.go
```diff
@@ -31,3 +31,42 @@ type MetricsBackendList struct {
...
+	Engine         string            `json:"engine"`
+	CooldownPeriod int               `json:"cooldownPeriod"`
+	Suspend        bool              `json:"suspend"`
+	MinNodes       int               `json:"minNodes"`
+	MaxNodes       int               `json:"maxNodes"`
+}
+
+// AutoscalingGroupStatus is the status for a autoscaling group
+type AutoscalingGroupStatus struct {
+	LastUpdatedAt int64 `json:"lastUpdatedAt"`
```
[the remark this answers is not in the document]
mattkelly: Makes sense to me as long as there's a comment

pkg/controller/autoscalinggroup.go
ashleyschuett: I think it would just default to random? should there also be a heuristic on an ASG?

pkg/apis/cerebral.containership.io/v1alpha1/types.go
```diff
@@ -31,3 +31,42 @@ type MetricsBackendList struct {
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   AutoscalingGroupSpec   `json:"spec"`
+	Status AutoscalingGroupStatus `json:"status"`
+}
+
+// AutoscalingGroupSpec is the spec for a autoscaling group
+type AutoscalingGroupSpec struct {
+	NodeSelector   map[string]string `json:"nodeSelector"`
```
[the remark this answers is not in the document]
nicktate: No, I think updating to match ‘nodeSelector’ makes sense.

pkg/controller/autoscalinggroup.go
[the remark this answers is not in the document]
nicktate: Throwing an error would imply the ASG controller crash loops? So the only way to resolve would be to update the CRs for the autoscaling groups not to select similar nodes?

That’s definitely the safest option. I’m not sure if I like that versus accepting the node as part of both ASGs and printing a warning message better. 

I think accepting it to just one or neither would be worse in my opinion.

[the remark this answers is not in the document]
ashleyschuett: I would be fine with both of these options. My preference is to queue both ASGs and print a warning that the node is being selected by both.

pkg/apis/cerebral.containership.io/v1alpha1/types.go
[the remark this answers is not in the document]
normanjoyner: It was set up to align with the `NodeSelector` field as defined in a `PodSpec` as [defined here](https://github.com/kubernetes/kubernetes/blob/bf7efc62722f8f077f9c1ff9bf1adc9d577f1ded/pkg/apis/core/types.go#L2545). I'm open to changing as we see fit.

pkg/autoscalingengine/autoscalingengine.go
```diff
@@ -0,0 +1,54 @@
...
+	Name() string
+	SetTargetNodeCount(nodeSelectorList labels.Selector, numNodes int, heuristic string) (bool, error)
+}
+
+// AutoscalingEngine keeps track of the engines that have been registered
+type AutoscalingEngine struct {
+	autoscalingEngines map[string]Interface
+}
+
+// New returns a new instances of AutoscalingEngine
```
normanjoyner: grammar

pkg/autoscalingengine/containership/containership.go
```diff
@@ -0,0 +1,32 @@
+package containership
+
+import (
+	"k8s.io/apimachinery/pkg/labels"
+
+	"github.com/containership/cluster-manager/pkg/log"
+)
+
+// Engine returns is an instance of the containership autoscaling engine
```
normanjoyner: grammar

pkg/autoscalingengine/containership/containership.go
```diff
@@ -0,0 +1,32 @@
...
+// TODO: implement the containership autoscaling engine
+type Engine struct {
+	name string
+}
+
+// NewAutoscalingEngine creates a new instances of the containership autoScaling
```
normanjoyner: grammar

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,376 @@
...
+	}
+
+	if autoScalingGroup.Spec.Suspend {
+		log.Infof("Autoscaling Group '%s' was queued but it is currently suspended.", autoScalingGroup.Name)
+		return nil
+	}
+
+	engineName := autoScalingGroup.Spec.Engine
+	if !agc.autoscalingengine.Exists(engineName) {
+		log.Errorf("The Autoscaling Engine Specified for the Autoscaling Group '%s' is not registered.", autoScalingGroup.Name)
```
normanjoyner: nit: capitalization of specified

pkg/apis/cerebral.containership.io/v1alpha1/types.go
```diff
@@ -31,3 +31,42 @@ type MetricsBackendList struct {
...
+	Policies       []string          `json:"policies"`
+	Engine         string            `json:"engine"`
+	CooldownPeriod int               `json:"cooldownPeriod"`
+	Suspend        bool              `json:"suspend"`
```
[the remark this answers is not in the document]
mattkelly: Based on the result of that discussion, this should be `Suspended`

pkg/controller/autoscalinggroup.go
normanjoyner: I think the AutoscalingGroup will need the heuristics regardless, as an AutoscalingPolicy should never know the implementation details of an AutoscalingEngine. I think this was overlooked in the RFC. @mattkelly is correct though that we will need to pass the proper heuristic here (up vs down). Defining it in the AutoscalingGroup does give us the ability to pass the proper heuristic here as you're alluding to @ashleyschuett

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,376 @@
...
+}
+
+// findNodesASG goes through each autoscaling group and checks to see if the
+// nodeSelector matches the node labels passed into the function
+// TODO: what if the user messes up and there's more than one ASG that matches,
+// should we find and enqueue all, or since we are assuming that each nodes should
+// only match one ASG is it okay to return the first one found?
+func findNodeASG(nodeLabels map[string]string, ags []*cerebralv1alpha1.AutoscalingGroup) *cerebralv1alpha1.AutoscalingGroup {
+	for _, autoScalingGroup := range ags {
+		// create selector abject from nodeSelector of ASG
```
normanjoyner: spelling

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,376 @@
...
+		agselectors := getNodesLabelSelector(autoScalingGroup.Spec.NodeSelector)
+
+		// check to see if the nodeSelector labels match the node labels that
+		// where passed in
```
normanjoyner: spelling

pkg/events/events.go
```diff
@@ -0,0 +1,8 @@
+package events
+
+const (
```
normanjoyner: should we also add the `ScaleIgnored` events?

normanjoyner: as well as `ScaleUpAlerted` and `ScaleDownAlerted`? or were these intentionally left for later implementation?

pkg/controller/autoscalinggroup.go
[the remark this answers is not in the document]
mattkelly: I was kinda thinking about this in the wrong context before. We can't return an error (well we could, but it'd be ignored). This is simply called from the `enqueue` function.

There should be no harm in queueing both ASGs. In a lot of cases, not queueing the ASGs here wouldn't change anything, because they'd be queued by some other event.

cmd/cerebral/cerebral.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/containership/cerebral/pkg/buildinfo"
+	cerebral "github.com/containership/cerebral/pkg/client/clientset/versioned"
+	cinformers "github.com/containership/cerebral/pkg/client/informers/externalversions"
+	"github.com/containership/cerebral/pkg/controller"
+
+	"github.com/containership/cluster-manager/pkg/log"
+)
+
+func main() {
+	log.Info("Starting Containership Autoscaling Engine Controller...")
```
mattkelly: `Starting Cerebral`

pkg/controller/autoscalinggroup.go
ashleyschuett: How should this look since the AutoscalingPolicy specifies a heuristic for both `ScaleUp` and `ScaleDown`. Should these properties also be added to the ASG? 

```
scaleup: 
 heuristic: random
scaledown:
  heuristic: cpu
```

cmd/autoscalingengine/autoscalingengine.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/containership/cluster-manager/pkg/log"
+)
+
+func main() {
+	log.Info("Starting Containership AutoScaling Engine Controller...")
+	log.Infof("Version: %s", buildinfo.String())
+	log.Infof("Go Version: %s", runtime.Version())
+
+	config := config.GetConfigOrDie()
```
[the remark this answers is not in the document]
ashleyschuett: Updated to use our own function for now. We can revisit pulling in a dependency later if there are other things we want from the controller-runtime package https://github.com/kubernetes-sigs/controller-runtime

pkg/events/events.go
ashleyschuett: I left them to be specified later. I'm not sure we want a package for this or if the package that uses them should define their own constants. Was kinda torn on it.

pkg/controller/autoscalinggroup.go
mattkelly: Very good question. I don't have a good answer. This could get complicated (or not, but it at least requires a lot more though), so I'm fine with filing a ticket and leaving it as-is for now.

I did some sleuthing through other controllers that would have a similar problem, such as the [volume attach/detach controller](https://github.com/mattkelly/kubernetes/blob/master/pkg/controller/volume/attachdetach/reconciler/reconciler.go). The basically keep a [local cache for pending status updates](https://github.com/mattkelly/kubernetes/blob/master/pkg/controller/volume/attachdetach/statusupdater/node_status_updater.go), which makes sense, but I imagine would still have issues in the case that the controller is restarted.

As an added bonus, the way they [unit test status update failures](https://github.com/mattkelly/kubernetes/blob/master/pkg/controller/volume/attachdetach/statusupdater/node_status_updater.go) (among other things) is interesting.

mattkelly: Specifically, can you create an issue and then reference it in a TODO here?

Gopkg.toml
```diff
@@ -12,8 +12,31 @@ required = [
...
+  name = "github.com/stretchr/testify"
+  version = "v1.2.2"
+
+[[constraint]]
+  name = "github.com/pkg/errors"
+  version = "0.8.0"
+
+[[constraint]]
+  name = "k8s.io/api"
+  version = "kubernetes-1.11.2"
```
mattkelly: Can't these be `1.12.3` again since that one dependency was removed?

ashleyschuett: 🤦‍♂️

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,398 @@
...
+	controllerName = "AutoscalingGroupController"
+
+	// this is the time delay between retries if a resource fails during sync
+	delayBetweenRequeues = 30 * time.Second
+
+	// number of times an autoscaling group will retry syncing
+	maxRequeues = 5
+
+	// the default scaling strategy to pass the autoscaling engine if one is not user defined
+	defaultAutoscalingStrategy = "random"
```
ashleyschuett: this should be "Default value provided by the associated `AutoscalingEngine`."

ashleyschuett: Should I add the CRD for it to this PR, or file a ticket to make sure it gets changed?

pkg/controller/autoscalinggroup.go
```diff
@@ -0,0 +1,398 @@
...
+
+	agLister clisters.AutoscalingGroupLister
+	agSynced cache.InformerSynced
+
+	workqueue workqueue.RateLimitingInterface
+	// recorder is an event recorder for recording Event resources to the
+	// Kubernetes API.
+	recorder record.EventRecorder
+
+	autoscalingengine *engine.AutoscalingEngine
```
mattkelly: `AutoscalingEngineRegistry`  probably would be a good name for this, but I'm not sold on it. We can leave as-is for now

pkg/controller/autoscalinggroup.go
mattkelly: Follow-up is fine with me

pkg/controller/autoscalinggroup_test.go
```diff
@@ -0,0 +1,230 @@
...
+		},
+		expected: []*cerebralv1alpha1.AutoscalingGroup{
+			emptyASG,
+			singleASG,
+			multipleLabelsASG,
+		},
+	},
+}
+
+func TestFindAGsMatchingNodeLabels(t *testing.T) {
```
mattkelly: Consistency with `AG` vs `ASG` through this entire changeset.

Issue #5, opened 2018-11-19
Initial implementation of Containership scaling engine with heuristic "random"


mattkelly:
See https://github.com/containership/internal-rfc/pull/3 for more information.

Initial implementation of the Containership scaling engine. This will always be in-tree, though other engines may eventually be out-of-tree.

Only the `random` heuristic should be implemented initially.

mattkelly:
Closed by https://github.com/containership/cerebral/pull/16.

PR #16, opened 2018-12-04
Add Containership Autoscaling Engine

ashleyschuett:
### Description

This adds the containership engine which will scale the node count to the numNodes that is passed into the `SetTargetNodeCount` function. 

* Addresses https://github.com/containership/cerebral/issues/5

 ### Testing

```
2018-12-04T13:53:42.123-0500	INFO	Current number of nodes in autoscaling group 'tester' : 1
2018-12-04T13:53:42.123-0500	INFO	Called SetTargetNodeCount
2018-12-04T13:53:42.123-0500	INFO	Requesting 'containership' autoscaling engine to set target nodes to 2
2018-12-04T13:53:42.522-0500	INFO	Event(v1.ObjectReference{Kind:"AutoscalingGroup", Namespace:"", Name:"tester", UID:"4be7d74e-f7f5-11e8-b076-1240cc647d73", APIVersion:"cerebral.containership.io/v1alpha1", ResourceVersion:"2365083", FieldPath:""}): type: 'Normal' reason: 'ScaledUp' Autoscaling group tester node count (1) was not within min and max bounds and was scaled up to 2
```

 #### Setup

This assumes your kubeconfig file is pointing to a CKE cluster, and that you have filled in the deploy examples for `autoscalinggroup.yaml`, and `autoscalingengine.yaml` with valid information.  

`kubectl apply -f deploy/crd`

`kubectl apply -f deploy/example`

`KUBECONFIG=$HOME/.kube/config go run cmd/cerebral/cerebral.go -v 9`

 #### Instructions

 ### Dependencies


Gopkg.toml
```diff
@@ -30,7 +30,11 @@ required = [
 
 [[constraint]]
   name = "k8s.io/client-go"
-  version = "kubernetes-1.12.3"
+  version = "9.0.0"
+
+[[constraint]]
+  branch = "master"
```
mattkelly: Please don't pull in branches. Instead just choose a commit for now.

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,89 @@
...
+	}
+
+	id, found := nodeSelectors[nodePoolIDLabel]
+	if !found {
+		return false, fmt.Errorf("could not get autoscaling group node pool id")
+	}
+
+	// TODO: is there a way to fake this to be tested?
+	log.Infof("Requesting 'containership' autoscaling engine to set target nodes to %d", numNodes)
+	_, err := cae.cloud.Provision().NodePools(cae.config.OrganizationID, cae.config.ClusterID).Scale(id, &req)
```
ashleyschuett: Since we are only implementing random for now I just added the Scale command. Should I be checking strategy and if not `random` do nothing? or is it okay to just assume this is correct for now?

cmd/cerebral/cerebral.go
```diff
@@ -45,8 +47,24 @@ func main() {
...
 
+	// Create autoscaling engine library
 	ae := autoscalingengine.New()
-	ae.Register(containership.NewAutoscalingEngine())
+
+	// get engine information by engine name
+	// TODO: This currently only works for containership and needs to be be changed
+	// to register other engines. This also assumes the Autoscaling Engine CR has
+	// the name of the engine you are registering
+	enginecr, err := cerebralclientset.Cerebral().AutoscalingEngines().Get("containership", metav1.GetOptions{})
```
mattkelly: We shouldn't be assuming that the engine exists on startup, and should instead use an informer on engines.

pkg/controller/autoscalinggroup.go
```diff
@@ -339,7 +339,7 @@ func (agc *AutoscalingGroupController) syncHandler(key string) error {
 
 	scaledUp := isScaleUpEvent(numNodes, desired)
 	strategy := getAutoscalingGroupStrategy(scaledUp, autoscalingGroup.Spec)
-	scaled, err := engine.SetTargetNodeCount(ns, desired, strategy)
+	scaled, err := engine.SetTargetNodeCount(autoscalingGroup.Spec.NodeSelector, desired, strategy)
```
ashleyschuett: I changed this to pass the map in, since when passing in type of `label.Selector` you can not get any key values out

cmd/cerebral/cerebral.go
```diff
@@ -45,8 +47,24 @@ func main() {
 	kubeInformerFactory := informers.NewSharedInformerFactory(kubeclientset, 30*time.Second)
 	cerebralInformerFactory := cinformers.NewSharedInformerFactory(cerebralclientset, 30*time.Second)
 
+	// Create autoscaling engine library
 	ae := autoscalingengine.New()
-	ae.Register(containership.NewAutoscalingEngine())
+
+	// get engine information by engine name
+	// TODO: This currently only works for containership and needs to be be changed
```
ashleyschuett: Is this something we want to address now, or wait for a follow up PR?

deploy/example/autoscalingengine.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: cerebral.containership.io/v1alpha1
+kind: AutoscalingEngine
+metadata:
+  name: containership
+spec:
+  type: containership
+  address: https://stage-provision.containership.io
+  configuration:
+    clusterToken: eyJ...
```
mattkelly: I think we need to put this into a secret before release.

pkg/apis/cerebral.containership.io/v1alpha1/types.go
```diff
@@ -126,3 +126,33 @@ type AutoscalingPolicyList struct {
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec AutoscalingEngineSpec `json:"spec"`
+}
+
+// AutoscalingEngineSpec describes the spec for the autoscaling engine
+type AutoscalingEngineSpec struct {
+	Address       string            `json:"address"`
+	Type          string            `json:"type"`
+	Configuration map[string]string `json:"configuration"`
```
mattkelly: I meant to write a TODO for the `MetricConfiguration` that it should be a `map[string]interface{}` and we should post-process the code gen to fix it (or some other solution). Maybe add TODOs now?

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,89 @@
...
+	"fmt"
+
+	cscloud "github.com/containership/csctl/cloud"
+	"github.com/containership/csctl/cloud/provision/types"
 
 	"github.com/containership/cluster-manager/pkg/log"
 )
 
+const (
+	nodePoolIDLabel = "containership.io/node-pool-id"
```
mattkelly: `LabelKey`

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,89 @@
...
+	log.Infof("Config: %+v", engine.config)
+
+	// TODO: anyway to test this without a real token?
+	cloudclientset, err := cscloud.New(cscloud.Config{
+		Token:            engine.config.ClusterToken,
+		ProvisionBaseURL: address,
+	})
+
+	if err != nil {
+		// TODO: should engine be return if theres an error? I feel like it shouldn't be
```
mattkelly: We definitely should not be returning the engine here.

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,89 @@
...
+
+	return engine, nil
 }
 
 // SetTargetNodeCount takes action to scale a target node pool
-func (cae *Engine) SetTargetNodeCount(nodeSelectorList labels.Selector, numNodes int, strategy string) (bool, error) {
+func (cae *Engine) SetTargetNodeCount(nodeSelectors map[string]string, numNodes int, strategy string) (bool, error) {
 	log.Info("Called SetTargetNodeCount")
+	if numNodes < 0 {
+		return false, fmt.Errorf("can not scale bellow 0")
```
mattkelly: Spelling.

pkg/autoscalingengine/containership/containership_test.go
```diff
@@ -0,0 +1,30 @@
...
+	assert.Equal(t, c.name, c.Name())
+}
+
+func TestSetTargetNodeCount(t *testing.T) {
+	c, err := NewAutoscalingEngine("address", map[string]string{})
+	assert.Error(t, err)
+
+	labels := make(map[string]string, 0)
+
+	// Test scaling below 0
```
mattkelly: Try to use messages in the `assert`s themselves over just commenting when it makes sense.

pkg/autoscalingengine/containership/containership_test.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestName(t *testing.T) {
+	c, err := NewAutoscalingEngine("address", map[string]string{})
+	assert.Error(t, err)
```
mattkelly: As alluded to previously, this isn't a valid test since accessing `c` if `err != nil` should be undefined behavior.

deploy/example/autoscalingengine.yaml
ashleyschuett: im not sure how we are going to do this as Custom Storage is not supported for CRDs. https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/#advanced-features-and-flexibility

ashleyschuett: Idk why im overthinking this. We would just add the secret to the pod/deployment running this. Should we add a key for the environment variable to look at?

cmd/cerebral/cerebral.go
```diff
@@ -85,3 +86,23 @@ func determineConfig() (*rest.Config, error) {
...
 }
+
+// registerContainershipEngine assumes a few things...
+// 1. It assumes that the containership autoscaling engine CR exists, if not cerebral fails
+// 2. Also, it assumes containership is the only autoscaling engine
+// 3. Finally if the containership autoscaling engine is unable to be created it
+//    assumes cerebral can't/shouldn't be used and fails.
+func registerContainershipEngine(cerebralclientset *cerebral.Clientset) {
+	// get engine information by engine name
+	enginecr, err := cerebralclientset.Cerebral().AutoscalingEngines().Get("containership", metav1.GetOptions{})
```
mattkelly: We shouldn't really assume any names and instead look for an engine with `Spec.Type = containership`. However, since the rest of this is so hardcoded to our setup with assumptions, maybe it's fine to leave as-is for now...

deploy/example/autoscalingengine.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: cerebral.containership.io/v1alpha1
+kind: AutoscalingEngine
+metadata:
+  name: containership
+spec:
+  type: containership
+  address: https://stage-provision.containership.io
```
mattkelly: According to the RFC, this belongs under `configuration`. I can imagine there will be weird engines for which `address` is not a thing.

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,85 @@
...
 		name: "containership",
 	}
+
+	engine.config = &cloudConfig{
+		ProvisionBaseURL: address,
+		ClusterToken:     configmap["clusterToken"],
+		OrganizationID:   configmap["organizationID"],
+		ClusterID:        configmap["clusterID"],
+	}
+	log.Infof("Config: %+v", engine.config)
```
mattkelly: You're logging a token at info level here. That's probably Not Great ™️

cmd/cerebral/cerebral.go
ashleyschuett: Yeah I was going back and forth on that. If I list the engines I would be looking for one with type "containership" and only register engines with that type. And since engines have to be unique by name, if multiple where registered it would only store the first one. It seemed weird to list and look for engines of type containership, but hard coding is also pretty awful.

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,84 @@
...
+
+	return engine, nil
 }
 
 // SetTargetNodeCount takes action to scale a target node pool
-func (cae *Engine) SetTargetNodeCount(nodeSelectorList labels.Selector, numNodes int, strategy string) (bool, error) {
-	log.Info("Called SetTargetNodeCount")
+func (cae *Engine) SetTargetNodeCount(nodeSelectors map[string]string, numNodes int, strategy string) (bool, error) {
+	if numNodes < 0 {
+		return false, fmt.Errorf("can not scale below 0")
```
mattkelly: `errors.New` instead, spelling: `cannot`

cmd/cerebral/cerebral.go
mattkelly: I'm not sure I quite follow the `if multiple where registered` part. Similar to the metrics backends, anything that uses them should reference them by name (which in the case of a Containership MB is an ID). We then use `type` to determine _how_ to talk to the MB.

It can be the same here. Two engines with the name `containership` can't exist in Kubernetes anyway.

deploy/crd/autoscaling-engine.yaml
```diff
@@ -0,0 +1,34 @@
...
+spec:
+  group: cerebral.containership.io
+  names:
+    kind: AutoscalingEngine
+    listKind: AutoscalingEngineList
+    singular: autoscalingengine
+    plural: autoscalingengines
+    shortNames:
+    - ase
+  # defined as Cluster scoped since AutoscalingEngine are universal to the cluster
```
mattkelly: `AutoscalingEngines` or just delete the comment IMO

deploy/example/autoscalingengine.yaml
mattkelly: I think that for the initial release, it wouldn't be the worst thing to put solely the token in a secret (or reference the existing one) and just pull it in via environment variable in the engine. It's asymmetric but at least correct for the token itself.

pkg/apis/cerebral.containership.io/v1alpha1/types.go
```diff
@@ -126,3 +126,33 @@ type AutoscalingPolicyList struct {
...
+type AutoscalingEngine struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec AutoscalingEngineSpec `json:"spec"`
+}
+
+// AutoscalingEngineSpec describes the spec for the autoscaling engine
+type AutoscalingEngineSpec struct {
+	Address       string            `json:"address"`
```
mattkelly: This belongs under `configuration`. I think I previously commented, but not sure where that went.

pkg/autoscalingengine/containership/containership.go
mattkelly: This is probably fine for now, but it would also be minimal effort to add a check on the strategy and 1) default to `random` if it's not set and 2) return an error if it's set but not `random`

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,84 @@
...
+		Count: &target,
+	}
+
+	id, found := nodeSelectors[nodePoolIDLabelKey]
+	if !found {
+		return false, fmt.Errorf("could not get autoscaling group node pool id")
+	}
+
+	// TODO: is there a way to fake this to be tested?
+	log.Infof("Requesting 'containership' autoscaling engine to set target nodes to %d", numNodes)
```
mattkelly: We're already in the engine here, so a better log message might be something like `'containership' autoscaling engine requesting Containership Cloud to set target nodes to ...`

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,84 @@
...
+	_, err := cae.cloud.Provision().NodePools(cae.config.OrganizationID, cae.config.ClusterID).Scale(id, &req)
+	if err != nil {
+		return false, err
```
mattkelly: `errors.Wrap`

pkg/autoscalingengine/registry.go
```diff
@@ -0,0 +1,56 @@
+package autoscalingengine
+
+import (
+	"fmt"
```
mattkelly: `errors` package

pkg/autoscalingengine/registry.go
```diff
@@ -0,0 +1,56 @@
...
+	"sync"
+
+	"github.com/containership/cluster-manager/pkg/log"
+)
+
+// RegistryInterface is an interface to a AutoscalingEngine registry.
+// See function comments for implementation limitations.
```
mattkelly: I don't think this comment makes sense here

pkg/autoscalingengine/registry.go
```diff
@@ -0,0 +1,56 @@
...
+type RegistryInterface interface {
+	Add(engine AutoscalingEngine)
+	Get(name string) (AutoscalingEngine, error)
+	IsRegistered(name string) bool
+}
+type registry struct {
+	sync.RWMutex
```
mattkelly: You have a mutex here that's not used. That means one of a few things:

1. We don't need a mutex, in which case this needs a comment explaining that it is not thread safe (and a TODO to make it thread safe one day if that's going to be needed)
2. We do need a mutex and should use it properly, keeping in mind that we're returning a interface in `Get` that will have no guarantees on validity outside of the function context so the onus is on the caller to behave appropriately (this makes sense for the MB registry and is documented appropriately, but does it make sense here?).

pkg/autoscalingengine/registry_test.go
```diff
@@ -30,44 +28,46 @@ func (te *TestEngine) Name() string {
 // End Test engine
 
 func TestRegister(t *testing.T) {
-	ae := New()
+	r := Registry()
```
mattkelly: You have multiple tests which are mutating global state. That seems bad. Maybe don't use the `Registry()` interface unless you explicitly clean up every time?

pkg/controller/autoscalinggroup.go
```diff
@@ -65,23 +65,22 @@ type AutoscalingGroupController struct {
 	// Kubernetes API.
 	recorder record.EventRecorder
 
-	autoscalingengine *engine.AutoscalingEngine
+	//autoscalingengine *engine.AutoscalingEngine
```
mattkelly: Please make sure to delete stale code

pkg/controller/autoscalinggroup.go
```diff
@@ -314,7 +313,7 @@ func (agc *AutoscalingGroupController) syncHandler(key string) error {
 	}
 
 	engineName := autoscalingGroup.Spec.Engine
-	if !agc.autoscalingengine.IsRegistered(engineName) {
+	if !autoscalingengine.Registry().IsRegistered(engineName) {
 		log.Errorf("The Autoscaling Engine specified for the Autoscaling Group '%s' is not registered.", autoscalingGroup.Name)
```
mattkelly: Why wouldn't we return this as an error here, instead of just dropping things?

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,28 +1,84 @@
...
-// TODO: implement the containership autoscaling engine
 type Engine struct {
-	name string
+	name   string
+	cloud  cscloud.Interface
+	config *cloudConfig
+}
+
+type cloudConfig struct {
+	Address        string `yaml:"address"`
```
mattkelly: Is there a reason to include the `yaml` annotations if they aren't needed?

cmd/cerebral/cerebral.go
ashleyschuett: Right now the autoscale engines are not being registered with the name of the CR that backs the Autoscaling engine, but rather the name that the engine assigns itself.

pkg/autoscalingengine/registry_test.go
ashleyschuett: oh duh. thank you.

cmd/cerebral/cerebral.go
```diff
@@ -85,3 +87,46 @@ func determineConfig() (*rest.Config, error) {
...
+
+	enginecrs, err := cerebralclientset.Cerebral().AutoscalingEngines().List(metav1.ListOptions{})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	for _, enginecr := range enginecrs.Items {
+		// only register engines of type containership
+		switch enginecr.Spec.Type {
+		case "containership", "Containership":
```
mattkelly: I think do:

```
switch {
    case strings.ToLower(enginecr.Spec.Type) == "containership"
```

Or just requiring it to be `containership` seems fine IMO since ideally this will be a strict enum one day anyway

cmd/cerebral/cerebral.go
```diff
@@ -85,3 +87,46 @@ func determineConfig() (*rest.Config, error) {
...
+		}
+
+	}
+
+	if !engineCreated {
+		log.Fatalf("No AutoscalingEngine of type containership found. Failed to create an autoscaling engine")
+	}
+}
+
+func initalizeAndRegisterContainershipEngine(enginecr cerebralv1alpha1.AutoscalingEngine) (bool, error) {
```
mattkelly: Spelling

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,32 +1,107 @@
...
+	if numNodes < 0 {
+		return false, errors.New("cannot scale below 0")
+	}
+
+	id, found := nodeSelectors[nodePoolIDLabelKey]
+	if !found {
+		return false, errors.New("could not get autoscaling group node pool ID")
+	}
+
+	log.Infof("Requesting 'containership' autoscaling engine requesting Containership Cloud to set target nodes %v to %d", nodeSelectors, numNodes)
```
mattkelly: You have `requesting` twice here

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,32 +1,107 @@
...
 
 // Name returns the name of the engine
 func (cae *Engine) Name() string {
 	return cae.name
 }
+
+// ScaleStrategyRandom take in the number of desired nodes for a node pool.
+// It then makes a request to Containership Cloud API to set the node pool to
+// the desired count
+func (cae *Engine) ScaleStrategyRandom(nodePoolID string, numNodes int) (bool, error) {
```
mattkelly: This can be unexported

pkg/autoscalingengine/registry.go
```diff
@@ -0,0 +1,54 @@
+package autoscalingengine
+
+import (
+	"github.com/containership/cluster-manager/pkg/log"
+
+	"github.com/pkg/errors"
+)
+
+// RegistryInterface is an interface to an AutoscalingEngine registry
+type RegistryInterface interface {
```
mattkelly: I think we should add a comment that this isn't thread safe, because I can see it being accessed by multiple threads (goroutines) in the (near) future

pkg/autoscalingengine/registry.go
```diff
@@ -0,0 +1,54 @@
...
+var reg = registry{
+	items: make(map[string]AutoscalingEngine),
+}
+
+// Registry returns the registry
+func Registry() RegistryInterface {
+	return &reg
+}
+
+// Add registers an instance of an AutoscalingEngine to the registry
```
mattkelly: Extend comment to explain that if the key already exists, this is a no-op (it doesn't overwrite it). (Or maybe overwriting is the correct thing to do?)

pkg/controller/autoscalinggroup.go
```diff
@@ -64,24 +64,21 @@ type AutoscalingGroupController struct {
...
-	ae *engine.AutoscalingEngine) *AutoscalingGroupController {
+	cInformerFactory cinformers.SharedInformerFactory) *AutoscalingGroupController {
 	rateLimiter := workqueue.NewItemExponentialFailureRateLimiter(delayBetweenRequeues, maxRequeues)
 
 	agc := &AutoscalingGroupController{
 		kubeclientset:     kubeclientset,
 		cerebralclientset: cerebralclientset,
 		workqueue:         workqueue.NewNamedRateLimitingQueue(rateLimiter, controllerName),
-		autoscalingengine: ae,
+		//autoscalingengine: ae,
```
mattkelly: Dead code

cmd/cerebral/cerebral.go
```diff
@@ -99,3 +102,46 @@ func determineConfig() (*rest.Config, error) {
...
+		// only register engines of type containership
+		switch strings.ToLower(enginecr.Spec.Type) {
+		case "containership":
+			engineCreated, err = initializeAndRegisterContainershipEngine(enginecr)
+			if err != nil {
+				log.Fatalf("Failed to create Containership autoscaling engine: %+v", err)
+			}
+
+		default:
+			log.Infof("Autoscaling Engine of type '%s' is unable to be registered", enginecr.Spec.Type)
```
mattkelly: `%q`

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,32 +1,107 @@
...
+
+type cloudConfig struct {
+	Address         string
+	TokenEnvVarName string
+	OrganizationID  string
+	ClusterID       string
 }
 
-// NewAutoscalingEngine creates a new instance of the containership autoScaling
+// NewAutoscalingEngine creates a new instance of the containership AutoScaling
```
mattkelly: `Autoscaling` but 🤷‍♂️

pkg/autoscalingengine/containership/containership.go
```diff
@@ -1,32 +1,107 @@
...
+			ClusterID:       configmap["clusterID"],
+		},
 	}
+
+	token := os.Getenv(engine.config.TokenEnvVarName)
+	if token == "" {
+		return nil, errors.New("unable to get Containership Cloud API cluster token")
+	}
+
+	// TODO: is there anyway to test this without a real token?
```
mattkelly: I've just been ignoring this for now; let's revisit soon

PR #17, opened 2018-12-04
Change configuration to be a string map of interfaces

ashleyschuett:
### Description

 This updates configuration keys to use type interface as their value. The addition of the struct with a DeepCopy() function was needed as a work around for code gen https://github.com/kubernetes/code-generator/issues/50

* None

 ### Testing

Was able to build without errors  

 #### Setup

 #### Instructions

 ### Dependencies


mattkelly:
CI issues look like flakes - rerunning.

PR #19, opened 2018-12-05
Add MetricsBackendController, fix some misc type issues

mattkelly:

 I'm breaking out some logical components for easier review, towards the goal of getting the entire metrics backend integration in shortly.

 - Add MetricsBackendController for managing clients for backends
 - Fix some type issues (see commit breakdown)

* None yet!

 ### Testing

 #### Setup

 - `telepresence` onto cluster (for now)

 #### Instructions

 - `KUBECONFIG=~/.kube/config LOG_LEVEL=DEBUG go run cmd/cerebral/cerebral.go`

I'm mostly testing on my full integration branch, but basically:

- `kubectl apply -f deploy/example/metricsbackend.yaml` - see backend created
- `kubectl delete -f deploy/example/metricsbackend.yaml` - see backend deleted
- etc

 ### Dependencies
* None


pkg/metrics/backend.go
```diff
@@ -9,5 +9,5 @@ type Backend interface {
 	// GetValue queries the backend and returns the raw numerical value of the
 	// requested metric (with the given configuration) for the given nodes
 	// at this point in time.
-	GetValue(metric string, configuration map[string]string, nodes []corev1.Node) (float64, error)
+	GetValue(metric string, configuration map[string]string, nodes []*corev1.Node) (float64, error)
```
mattkelly: This change is required in order to pass nodes from listers rather than from the API response. @ashleyschuett hinted at this before.

pkg/controller/metrics_backend.go
```diff
@@ -0,0 +1,276 @@
...
+		return err
+	}
+
+	// If the instantiated metrics backend already exists, let's just clean it up
+	// and replace it instead of trying to update it. By deleting it here, we'll
+	// cause anyone using the backend to see that it's no longer around and thus
+	// backoff into a retry loop.
+	if _, err := metrics.Registry().Get(name); err == nil {
+		log.Infof("%s: backend for %q already exists - it will be replaced", metricsBackendControllerName, name)
+		metrics.Registry().Delete(name)
```
mattkelly: This is pretty much the crux of why it's architected this way. By doing this, we're essentially broadcasting to all pollers "hey the backend changed and your client is no longer good; please retry".

An alternative approach would be to have each backend poller own its own client. I don't see a particular reason for this. In this case, we would have to further overload the metrics controller to listen to MetricsBackends, etc. We'd then have to trace the chain of MB -->  ASP --> ASG to queue the ASG, and then in the ASG `syncHandler` determine if the client was still valid.

There are probably other clean ways to do it, but this felt like a natural separation of concerns to me. Always open to input though!

ashleyschuett: I don't love this design as we have talked about. I'm not sure having a controller is the right way to go about registering things, however don't have any ideas right now and don't think its necessarily bad but I think we should have further discussions about it at a later time.

mattkelly: I would be interested in either a comment on this PR or a ticket further elaborating your thoughts on why this is not an optimal approach.

PR #21, opened 2018-12-07
change AdjustmentValue type int -> float64

ashleyschuett:
### Description

Change adjustment value since it can be an absolute value, or a percent and should not be rounded. 

This will also need to be updated in `cluster-manager` once merged

 ### Testing

 #### Setup

 #### Instructions

 ### Dependencies


mattkelly:
Note to self: do something about the case where `AdjustmentType = absolute` but a whole number is not specified.

mattkelly:
You can just self-merge when CI passes; I'm 😴🛌⚰️

PR #20, opened 2018-12-06
Initial implementation of metrics polling

mattkelly:
#### What does this pull request accomplish?

 Splitting off more work for easy review.
 This is an initial implementation of metrics polling (scale up only - needs refactoring to add down).

 It integrates entirely up through the point where scales should actually be requested.
 This is because I believe we should add a scale request manager (goroutine) just before the actual scaling engine that should:

 - Receive incoming scale requests that include: ASG name, scale up/down(/to?), adjustmentType, adjustmentValue
 - Get the ASG from cache so it has latest
 - Get current node count via node selector (i.e. re-list nodes in order to have latest)
 - Calculate desired node count using scale up/down/to, adjustmentType, adjustmentValue
 - Decide whether to actually scale based on cooldown, current node count, and desired node count
 - Actually scale by calling engine
 - Update ASG status if needed
   - (As a near-future optimization, it can/should block incoming scale requests until it successfully updates the ASG status)

 #### What issue(s) does this fix?
* None yet

 #### Additional Considerations

This is totally unoptimized at the moment. Basically, poll loops are handled by spinning of a goroutine for each ASG. This fans out into N more goroutines, owned by the ASG polling manager, to poll for each of the policies.

If anything changes with an ASG or ASP referenced by an ASG, that polling manager is cleaned up entirely and a new one is created and spun up. This can obviously be optimized but it doesn't feel important right now - how often are ASG and ASP resources going to be changing?

 ### Testing

- Starting up with a few ASPs added to an ASG results in proper number of pollers chugging away:

```
2018-12-06T17:17:12.521-0500    INFO    Starting Cerebral...
2018-12-06T17:17:12.521-0500    INFO    Version: unknown (unknown) built @ 1969-12-31 19:00:00 -0500 EST
2018-12-06T17:17:12.521-0500    INFO    Go Version: go1.11.2
2018-12-06T17:17:12.521-0500    INFO    Using out of cluster k8s config: /Users/matt/.kube/config
2018-12-06T17:17:12.522-0500    INFO    Registered Autoscaling Engine "containership"
2018-12-06T17:17:12.522-0500    INFO    Setting up event handlers
2018-12-06T17:17:12.523-0500    INFO    MetricsBackendController: setting up event handlers
2018-12-06T17:17:12.523-0500    INFO    MetricsController: setting up event handlers
2018-12-06T17:17:12.523-0500    INFO    Starting MetricsController
2018-12-06T17:17:12.523-0500    INFO    Starting controller
[...]

PR #23, opened 2018-12-07
Initial implementation of scale manager

### LINKED ISSUE #1: Initial scaffolding, CRDs, and generated code
Initial repo structure should include:

- Travis integration
- codecov.io integration
- Initial Makefile
- Initial `dep` stuff

Initial CRD work should include:
- CRDs
- Scripts to generate and verify code, exposed as Makefile targets

### LINKED ISSUE #4: Initial integration of metrics backend to trigger scale events
See https://github.com/containership/internal-rfc/pull/3 for more details.

Perform the initial integration of the metrics backend within the AutoScalingController. This integration should be up to the point where the ScalingEngine API is called, which can be a no-op for now.

Blocked by https://github.com/containership/cerebral/issues/2. (This technically doesn't have to be blocked by that since it should be modular enough, but marking as blocked for now anyway just to enforce some priority.)

mattkelly:

This adds the scale manager, which acts as the final stage before the actual engine is called. More context is available in the comments and in the description of #20.

This is a WIP. TODO:

- [x] Make the `AutoscalingGroupController` make requests to the scale manager instead of hitting the engine directly
- ~[ ] More tests~ will do in a future PR

* Closes #4 

 ### Testing

 #### Setup

- Provision a cluster on DO
- Apply an ASG with appropriate `containership/node-pool-id` node selector to select worker node pool

 #### Instructions

- Apply an ASP such that the current values returned by Prometheus will result in alerting
- See that a scale alert is triggered after the sample period is reached
- See that subsequent scale requests are ignored due to the cooldown being in effect

 ### Dependencies
* None

### DIFF
--- a/pkg/controller/metric_poller.go
+++ b/pkg/controller/metric_poller.go
@@ -23,14 +23,23 @@ type alertState struct {
 	startTime time.Time
 }
 
+var nowFunc = time.Now
+
+func (a *alertState) start() {
+	a.startTime = nowFunc()
+	a.active = true
+}
+
+func (a *alertState) shouldFire(samplePeriod time.Duration) bool {
+	return a.active && nowFunc().Sub(a.startTime) >= samplePeriod
+}
+
 func newMetricPoller(asp *v1alpha1.AutoscalingPolicy, nodeSelector map[string]string) metricPoller {
 	return metricPoller{
 		asp:          asp,
 		nodeSelector: nodeSelector,
 	}
 }
-
-var nowFunc = time.Now
 
 func (p metricPoller) run(wg *sync.WaitGroup, alertCh chan<- alert, stopCh <-chan struct{}) {
 	defer wg.Done()
@@ -42,8 +51,8 @@ func (p metricPoller) run(wg *sync.WaitGroup, alertCh chan<- alert, stopCh <-cha
 	metric := p.asp.Spec.Metric
 	metricConfig := p.asp.Spec.MetricConfiguration
 
-	var upAlert = &alertState{}
-	var downAlert = &alertState{}
+	upAlert := &alertState{}
+	downAlert := &alertState{}
 
 	ticker := time.NewTicker(pollInterval)
 	for {
@@ -65,23 +74,16 @@ func (p metricPoller) run(wg *sync.WaitGroup, alertCh chan<- alert, stopCh <-cha
 
 			log.Debugf("Poller for ASP %q got value %f", policyName, val)
 
-			// Check for scale up alerts
-			shouldAlertUp := evaluatePolicyConfiguration(
-				p.asp.Spec.ScalingPolicy.ScaleUp, upAlert, samplePeriod, val)
-			if shouldAlertUp {
-				alertCh <- alert{
-					aspName:   p.asp.ObjectMeta.Name,
-					direction: "up",
-				}
+			// Scale up alerts
+			upConfig := p.asp.Spec.ScalingPolicy.ScaleUp
+			if policyConfigurationShouldFireAlert(upConfig, upAlert, samplePeriod, val) {
+				p.fireAlert(alertCh, upConfig, scaleDirectionUp)
 			}
 
-			shouldAlertDown := evaluatePolicyConfiguration(
-				p.asp.Spec.ScalingPolicy.ScaleDown, downAlert, samplePeriod, val)
-			if shouldAlertDown {
-				alertCh <- alert{
-					aspName:   p.asp.ObjectMeta.Name,
-					direction: "down",
-				}
+			// Scale down alerts
+			downConfig := p.asp.Spec.ScalingPolicy.ScaleDown
+			if policyConfigurationShouldFireAlert(downConfig, downAlert, samplePeriod, val) {
+				p.fireAlert(alertCh, downConfig, scaleDirectionDown)
 			}
 
 		case <-stopCh:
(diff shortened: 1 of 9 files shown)


pkg/controller/poll_manager.go
```diff
@@ -17,19 +17,22 @@ type pollManager struct {
 	// Key is ASP name
```
mattkelly: (file comment) All of this stuff should probably live in its own package and generally be organized better, when we have some spare time

pkg/controller/scale_manager.go
```diff
@@ -0,0 +1,240 @@
...
+			asg, err := m.asgLister.Get(req.asgName)
+			if err != nil {
+				if kubeerrors.IsNotFound(err) {
+					log.Infof("%s: AutoscalingGroup %q was deleted - ignoring scale request", scaleManagerName, req.asgName)
+					req.errCh <- nil
+					break
+				}
+
+				req.errCh <- errors.Wrapf(err, "getting AutoscalingGroup %q to scale", req.asgName)
+				break
```
mattkelly: I'm not a huuuuuge fan of this pattern for a few reasons; something to think further about

pkg/controller/scale_manager.go
```diff
@@ -0,0 +1,240 @@
...
+// to the ScaleManager
+func (m *ScaleManager) ScaleRequestChan() chan<- ScaleRequest {
+	return m.scaleRequestCh
+}
+
+// Run runs the ScaleManager. It should never return under normal conditions.
+func (m *ScaleManager) Run(stopCh <-chan struct{}) error {
+	for {
+		select {
+		case req := <-m.scaleRequestCh:
```
mattkelly: This `case` is too long. Can it 1) be shorter 2) be tested more easily

pkg/controller/scale_manager_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+type fitWithinBoundsTest struct {
+	name     string
+	val      int
+	min      int
+	max      int
+	expected int
+}
+
+var fitWithinBoundsTests = []fitWithinBoundsTest{
```
mattkelly: I borrowed and renamed this stuff from a different file; thanks @ashleyschuett

pkg/controller/metric_poller.go
```diff
@@ -42,8 +42,8 @@ func (p metricPoller) run(wg *sync.WaitGroup, alertCh chan<- alert, stopCh <-cha
 	metric := p.asp.Spec.Metric
 	metricConfig := p.asp.Spec.MetricConfiguration
 
-	var upAlert = &alertState{}
-	var downAlert = &alertState{}
+	upAlert := &alertState{}
```
ashleyschuett: just curious why you decided to change the convention?

pkg/controller/metric_poller.go
```diff
@@ -23,15 +23,15 @@ type alertState struct {
 	startTime time.Time
 }
 
+var nowFunc = time.Now
```
mattkelly: Also not a huge fan of this for testing purposes. Something else to rethink in the near future

pkg/controller/metric_poller.go
```diff
@@ -66,21 +66,29 @@ func (p metricPoller) run(wg *sync.WaitGroup, alertCh chan<- alert, stopCh <-cha
 			log.Debugf("Poller for ASP %q got value %f", policyName, val)
 
 			// Check for scale up alerts
-			shouldAlertUp := evaluatePolicyConfiguration(
-				p.asp.Spec.ScalingPolicy.ScaleUp, upAlert, samplePeriod, val)
+			upConfig := p.asp.Spec.ScalingPolicy.ScaleUp
```
ashleyschuett: could this through line 80 be turned into a function instead of having the same thing twice for scale down?

ashleyschuett: something like this? 
```
checkForAlert(p.asp.Spec.ScalingPolicy.ScaleUp)

func checkForAlert(config,... ) {
 shouldAlert := evaluatePolicyConfiguration(config, upAlert, samplePeriod, val)
 if shouldAlert {
  adjustmentType, _ := adjustmentTypeFromString(config.AdjustmentType)
  alertCh <- alert{
    aspName:         p.asp.ObjectMeta.Name,
    direction:       scaleDirectionUp,
    adjustmentType:  adjustmentType,
    adjustmentValue: config.AdjustmentValue,
  }
 }


}
```

pkg/controller/scale_manager.go
```diff
@@ -0,0 +1,240 @@
...
+
+func isCoolingDown(asg *cerebralv1alpha1.AutoscalingGroup) bool {
+	if asg.Status.LastUpdatedAt == 0 {
+		return false
+	}
+
+	return (nowFunc().Unix() - asg.Status.LastUpdatedAt) < int64(asg.Spec.CooldownPeriod)
+}
+
+func (m *ScaleManager) updateAutoscalingGroupStatus(autoscalingGroup *cerebralv1alpha1.AutoscalingGroup) error {
```
ashleyschuett: does it make sense to share this https://github.com/containership/cerebral/blob/master/pkg/controller/autoscalinggroup.go#L357-L365

pkg/controller/metric_poller.go
mattkelly: Just succinctness / consistency with elsewhere. Not sure why I had done it this way.

pkg/controller/scale_manager.go
mattkelly: I can refactor it to not have a receiver and instead take the `clientset` as an arg. Totally forgot about this.

pkg/controller/scale_manager.go
```diff
@@ -0,0 +1,243 @@
...
+			if err != nil {
+				req.errCh <- errors.Wrapf(err, "listing nodes for AutoscalingGroup %q", req.asgName)
+				break
+			}
+
+			currNodeCount := len(nodes)
+			targetNodeCount := calculateTargetNodeCount(currNodeCount, asg.Spec.MinNodes, asg.Spec.MaxNodes,
+				req.direction, req.adjustmentType, req.adjustmentValue)
+
+			scaled, err := engine.SetTargetNodeCount(asg.Spec.NodeSelector, targetNodeCount, "random")
```
mattkelly: `random` should no longer be hardcoded 🤦‍♂️

pkg/controller/metric_poller.go
mattkelly: Nice catch. For some reason my brain was exploding trying to think about this. I did _almost_ what you said; hopefully the result is ok.

PR #24, opened 2018-12-08
add default scaling strategy to engine

ashleyschuett:
### Description

The autoscaling engine should be able to have a default strategy for scaling so that an autoscaling strategy is not defined on the autoscaling group it uses the engines default strategy in scale cases. 

The actual autoscaling engine needs to be changed to use this new property but I am waitron for PR #23 before refactoring that. I just wanted to throw this PR up so I don't forget to do it. 

 ### Testing

 #### Setup

 #### Instructions

### DIFF
--- a/deploy/crd/autoscaling-engine.yaml
+++ b/deploy/crd/autoscaling-engine.yaml
@@ -26,5 +26,7 @@ spec:
           properties:
             type:
               type: string
+            defaultScalingStrategy:
+              type: string  
             configuration:
               type: object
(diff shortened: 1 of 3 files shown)


mattkelly:
Attempt #3 at reviewing: this does belong to the engine. Do we need to specify it explicitly in the CR though, or should it just be internal to the engine? I believe we previously discussed the latter.

ashleyschuett:
When I went to refactor to accommodate this change I realized it does have to be added to the Engine interface, so it seems like putting it in the CR might be overkill, unless we want it to be configurable at a later date

ashleyschuett:
Because scaling strategy is configurable on the ASG which is used for both ASG and ASE scaling this field is no longer needed. 

Cerebral will instead pass `""` to the engines `SetTargetNodeCount()` if there is no strategy specified for the current action.

PR #25, opened 2018-12-09
Default to node_cpu, fix issue where using stale config type

mattkelly:

Default to node_cpu, fix issue where using stale config type

Since many things assume Containership today, let's make our lives easier moving forward and just default to the Containership-supported `node_cpu` metric name we don't have to specify it in the config (which should go away anyway - see TODO).

Also fixes an issue where we were using a stale type I defined and never deleted.
This is a no-op change as the struct was exactly equal to the other struct type.

* Fixes containership/cloud.provision#508

 ### Testing

Run with an ASP that does not have the `cpuMetricName` key and see that `node_cpu` is defaulted.

 ### Dependencies
* None


mattkelly:
I didn't mean to close this 🤷‍♂️

PR #34, opened 2018-12-10
update LastUpdatedAt field to be human readable

### LINKED ISSUE #32: AutoscalingGroup lastUpdatedAt timestamp should be human readable
The `lastUpdatedAt` timestamp is not human readable.

It would be nice if it were human readable, and in the same format Kubernetes uses for things like `creationTimestamp`, etc.

#### Additional Context
Current implementation:
```
status:
  lastUpdatedAt: 1544394276
```

Desired implementation:
```
status:
  lastUpdatedAt: 2018-12-09T22:24:36Z
```

ashleyschuett:
### Description

Changed the `LastUpdateAt` field under `Status` of an ASG to be of type `metav1.Time`. This allows us to use a human readable timestamp that is shown in the status object of the CR, as well as allows the field to be of type Time. Using the golang type of `time.Time` breaks codegen. 

This PR also removed the unused function `updateAutoscalingGroupStatus` from autoscaling group as all status updates are now done in the scale manager. 

* Fixes #32 

 ### Testing

Updated golang tests to use new timestamp and they passed. 

 #### Setup

 #### Instructions

 ### Dependencies


pkg/controller/scale_manager_test.go
```diff
@@ -208,27 +208,27 @@ func TestIsCoolingDown(t *testing.T) {
...
-		},
+		Status: v1alpha1.AutoscalingGroupStatus{},
 	}
 
-	setTime(2)
+	// Special case: a scale has never been triggered and thus LastUpdatedAt is unset
+	setTime(0) // doesn't matter but just so it's a known value
+	assert.False(t, isCoolingDown(asg), "unset LastUpdatedAt means not cooling down")
+
+	now := metav1.Now()
```
mattkelly: I'm not a huge fan of depending on using real time in tests, but this seems fine

ashleyschuett: I can instead change it to 
```
now := time.Date(2009, time.November, 10, 23, 0, 0, 0, time.UTC)
asg.Status.LastUpdatedAt = metav1.Time{
  Time: now,
}
```

mattkelly: I would be a fan of that if you're up for it

PR #42, opened 2018-12-11
Expand on initial README

mattkelly:

Expand on initial README

The link to the `docs` directory is currently broken.
I intend to add it in the next PR.


README.md
```diff
@@ -10,5 +10,41 @@ Cerebral is a Kubernetes cluster autoscaler with pluggable metrics backends and
...
+Cerebral takes a more generic, flexible, and powerful approach to autoscaling by integrating with existing metric backends as input, and in turn, triggering pluggable actions in response to scaling events.
+
+# Learn More
+
+Please refer to our [documentation](/docs) for more information on building, configuring, and running Cerebral.
+
+# Project Status
+
+This project is in alpha.
+There may be backwards-incompatible breaking changes made as we continue to expand the project and integrated user feedback.
```
nicktate: This reads a little strange imo, I almost never se backwards-**in**compatible. Might be worth just dropping and saying `breaking changes`

README.md
```diff
@@ -10,5 +10,41 @@ Cerebral is a Kubernetes cluster autoscaler with pluggable metrics backends and
...
+
+Manually scaling nodes in a Kubernetes cluster is not feasible given the largely dynamic nature of web infrastructure; thus, automation is needed to assist operators in these tasks.
+With the increased importance placed on monitoring and observability in modern infrastructure, operators should be able to easily take action on the metrics they are collecting.
+
+## Alternatives
+
+Another tool for autoscaling is the [Kubernetes Autoscaler](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler).
+It is a prerequisite that the integrated providers support Autoscaling Groups (ASGs), a feature which many cloud providers do not have.
+
+Additionally, the method by which scaling occurs is naïve; often triggering events too late to be rendered useful.
```
nicktate: may be worth referencing: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md%23how-can-i-configure-overprovisioning-with-cluster-autoscaler

That is their "work around" but it runs into the same problems they explicitly stated they were avoiding and is much less flexible than usage based metrics.

README.md
mattkelly: Updated

README.md
mattkelly: I'll look into adding this as part of a near-future PR

PR #43, opened 2018-12-17
Refactor AutoscalingEngine to be dynamic

normanjoyner:
### Description

Refactors the autoscaling engine code such that Cerebral no longer requires the Containership engine to operate. `AutoscalingEngine` sync logic has been moved to its own controller, the engine registry was refactored, and the autoscaling package structure was updated to accommodate future engines. Finally, the polling manager was made aware of the autoscaling engine state, and will never start if the engine does not exist, and similarly, will shut down when the engine is removed.

N/A

N/A

 ### Testing

 #### Setup
Launched cluster with Cerebral installed and updated deployment to use custom build with requested changes.

 #### Instructions
Tested a variety of cases:
- [x] Start Cerebral with multiple ASGs with associated ASPs, Prometheus backend and Containership engine. Everything operated as expected.
- [x] Start Cerebral with multiple ASGs with associated ASPs, Prometheus backend and no engine. As expected, metric polling never started.
- [x] Add Containership engine to case above and observed metric polling start for each associated ASG.
- [x] Deleted Containership engine from case above and observed metric polling stop for each associated ASG.

 ### Dependencies
N/A


pkg/autoscaling/engines/containership/containership.go
```diff
@@ -31,41 +30,61 @@ type cloudConfig struct {
...
-	engine := &Engine{
-		name: e.Name,
-		config: &cloudConfig{
-			Address:         configmap["address"],
-			TokenEnvVarName: configmap["tokenEnvVarName"],
-			OrganizationID:  configmap["organizationID"],
-			ClusterID:       configmap["clusterID"],
-		},
+// NewClient creates a new instance of the containership AutoScaling Engine, or an error
+func NewClient(name string, address string, tokenEnvVarName string, clusterID string, organizationID string) (autoscaling.Engine, error) {
```
mattkelly: I might consider passing an options struct instead of all of these args

pkg/controller/metrics.go
```diff
@@ -120,12 +124,28 @@ func NewMetrics(kubeclientset kubernetes.Interface,
 		DeleteFunc: c.enqueueASGsForAutoscalingPolicy,
 	})
 
+	aseInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
```
ashleyschuett: Do we need to enqueue the ASG for AE updates or deletes, or is it fine to catch it later in the autoscaling engine? I would think it makes sense to process ASGs on AE adds incase some scaling actions failed due to the engine not existing, but not sure it makes sense on the other two cases.

pkg/controller/autoscaling_engine.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+// enqueueAutoscalingEngine enqueues an AutoscalingEngine object
+func (c *AutoscalingEngineController) enqueueAutoscalingEngine(obj interface{}) {
+	var key string
+	var err error
+	if key, err = cache.MetaNamespaceKeyFunc(obj); err != nil {
+		log.Error("Error enqueueing AutoscalingEngine: %s", err)
+		return
+	}
```
ashleyschuett: nit: new line

pkg/controller/autoscaling_engine.go
```diff
@@ -0,0 +1,264 @@
...
+	autoscaling.Registry().Put(name, client)
+	log.Infof("Engine %q instantiated successfully", name)
+
+	return nil
+}
+
+// insantiateEngine instantiates a new engine for the given AutoscalingEngine.
+// It should be the only function that knows how to instantiate a particular engine type.
+func (c *AutoscalingEngineController) instantiateEngine(engine *cerebralv1alpha1.AutoscalingEngine) (autoscaling.Engine, error) {
```
ashleyschuett: this seems like it should not be here, as it is engine specific code I would expect it to be in the logic when creating a new engine.

pkg/controller/metrics.go
```diff
@@ -306,6 +349,16 @@ func (c *MetricsController) syncHandler(key string) error {
 		return nil
 	}
 
+	_, err = c.aseLister.Get(asg.Spec.Engine)
+	if err != nil {
+		if kubeerrors.IsNotFound(err) {
+			log.Warnf("%s: No valid engines found for AutoscalingGroup %q - skipping", metricsControllerName, asgName)
+			return nil
+		} else {
+			return err
```
ashleyschuett: This needs taken out of the else

pkg/controller/autoscaling_engine.go
```diff
@@ -0,0 +1,264 @@
...
+	if _, err := autoscaling.Registry().Get(name); err == nil {
+		log.Infof("%s: engine for %q already exists - it will be replaced", autoscalingEngineControllerName, name)
+		autoscaling.Registry().Delete(name)
+	}
+
+	log.Infof("Instantiating engine client for AutoscalingEngine %q", name)
+	client, err := c.instantiateEngine(engine)
+	if err != nil {
+		return errors.Wrapf(err, "instantiating engine client for AutoscalingEngine %q", name)
+	}
```
ashleyschuett: nit: new line

pkg/autoscaling/engines/containership/containership.go
ashleyschuett: I agree all of these values should be in the `CR.configuration` object and be passed in as such

pkg/autoscaling/engines/containership/containership.go
```diff
@@ -31,41 +30,61 @@ type cloudConfig struct {
...
 
-	return engine, nil
+// Name returns the name of the engine
+func (cae Engine) Name() string {
+	return cae.name
 }
 
 // SetTargetNodeCount takes action to scale a target node pool
-func (cae *Engine) SetTargetNodeCount(nodeSelectors map[string]string, numNodes int, strategy string) (bool, error) {
+func (cae Engine) SetTargetNodeCount(nodeSelectors map[string]string, numNodes int, strategy string) (bool, error) {
```
mattkelly: nit: it'd probably be more consistent to just call the receiver `e` instead of `cae`

pkg/autoscaling/engines/containership/containership_test.go
```diff
@@ -22,8 +20,9 @@ func fakeAutoscalingEngine() *Engine {
 	}
 }
 
-func TestNewAutoscalingEngine(t *testing.T) {
-	_, err := NewAutoscalingEngine(cerebralv1alpha1.AutoscalingEngine{})
+func TestNewClient(t *testing.T) {
+	c := fakeAutoscalingEngine()
+	_, err := NewClient(c.name, c.config.Address, c.config.TokenEnvVarName, c.config.ClusterID, c.config.OrganizationID)
 	assert.Error(t, err)
```
mattkelly: Didn't catch this before - what is this testing? Why should it error? I'm guessing because the token won't be available from the environment. But this currently isn't a useful test without any context and without a corresponding positive case IMO.

pkg/controller/autoscaling_engine.go
```diff
@@ -0,0 +1,264 @@
...
+	switch engine.Spec.Type {
+	case "containership":
+		var name, address, tokenEnvVarName, clusterID, organizationID string
+		var ok bool
+
+		name = engine.Name
+
+		address, ok = engine.Spec.Configuration["address"]
+		if !ok {
+			return nil, errors.New("Containership engine requires address in configuration")
```
mattkelly: All of this checking is redundant with the checking performed in `NewClient()`, so I'd be ok with commenting that and simply `errors.Wrap`ing the error returned from `containership.NewClient()`, i.e.:

```
err = containership.NewClient(name, address, tokenEnvVarName, clusterID, organizationID)
if err != nil {
	return errors.Wrap(err, "constructing new containership engine")
}

return nil
```

pkg/autoscaling/engine.go
```diff
@@ -0,0 +1,7 @@
+package autoscaling
+
+// Engine specifies the functions that an Engine must implement
+type Engine interface {
+	Name() string
```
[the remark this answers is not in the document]
normanjoyner: I had originally removed it, but did find it useful in the one place [we are using it](https://github.com/containership/cerebral/blob/6339097e62b26ee75fe88ed28bd96147e8c34005/pkg/autoscalingengine/containership/containership.go#L78), so added it back.

normanjoyner:
@mattkelly @ashleyschuett updated with comment fixes. I'm happy to squash once there is approval. I will continue to add commits for readability until that time.

pkg/autoscaling/engines/containership/containership_test.go
```diff
@@ -22,9 +21,23 @@ func fakeAutoscalingEngine() *Engine {
...
+		"clusterID":       "cluster-uuid",
+	}
+
+	c, err := NewClient(name, configuration)
+	assert.Error(t, err, "Testing that an error is returned when the token environment variable is not defined")
+
+	os.Setenv(configuration["tokenEnvVarName"], "token")
+	c, err = NewClient(name, configuration)
+	assert.Equal(t, nil, err, "Testing that error is nil when client is successfully created")
+	assert.NotEqual(t, nil, c, "Testing that client is not nil when successfully created")
```
mattkelly: `assert.NotNil`

pkg/autoscaling/engines/containership/containership_test.go
```diff
@@ -22,9 +21,23 @@ func fakeAutoscalingEngine() *Engine {
...
+		"organizationID":  "organization-uuid",
+		"clusterID":       "cluster-uuid",
+	}
+
+	c, err := NewClient(name, configuration)
+	assert.Error(t, err, "Testing that an error is returned when the token environment variable is not defined")
+
+	os.Setenv(configuration["tokenEnvVarName"], "token")
+	c, err = NewClient(name, configuration)
+	assert.Equal(t, nil, err, "Testing that error is nil when client is successfully created")
```
mattkelly: `assert.NoError`

pkg/autoscaling/engines/containership/containership_test.go
```diff
@@ -22,9 +21,23 @@ func fakeAutoscalingEngine() *Engine {
...
+		"address":         "https://provision-test.containership.io",
+		"tokenEnvVarName": "TOKEN_ENV_VAR",
+		"organizationID":  "organization-uuid",
+		"clusterID":       "cluster-uuid",
+	}
+
+	c, err := NewClient(name, configuration)
+	assert.Error(t, err, "Testing that an error is returned when the token environment variable is not defined")
+
+	os.Setenv(configuration["tokenEnvVarName"], "token")
```
mattkelly: I'm not a fan of using the real `os.*env` functions for tests. Instead, we should e.g. wrap the `os` calls in some interface and mock that out.

I think that's out of scope for this PR and am fine with leaving this as-is for now.

pkg/controller/metrics.go
```diff
@@ -120,12 +124,25 @@ func NewMetrics(kubeclientset kubernetes.Interface,
 		DeleteFunc: c.enqueueASGsForAutoscalingPolicy,
 	})
 
+	// Explicitly ignore AutoscalingEngine updates since we only care about
```
mattkelly: I'm not sure I agree with this. Why wouldn't we want to just also listen to updates and handle appropriately, given that it's zero extra effort to make updates work properly (because the logic is already implemented [here](https://github.com/containership/cerebral/pull/43/files#diff-787a677f75b6a5b025031f9738e63317R210) as borrowed from the metrics side)? Why wouldn't I want the updated configuration to be picked up automatically? That would be consistent with how the metrics backend works.

ashleyschuett: the updates are listened to in the autoscaling engine controller and picked up and updated, why would the metrics backend care about the updates? when It gets to the scale manager its going to grab the engine which will have the updates and use them? unless im misunderstanding how the engine is being interacted with

mattkelly: As discussed over Slack since I stepped away, I was lost and thought this was in the engine controller. Nothing to do here.

PR #55, opened 2018-12-20
scale manager tests

### LINKED ISSUE #44: Unit tests for scale manager
We need more tests for the scale manager, since it's critical to the functionality of the system.

In particular, functions [like this](https://github.com/containership/cerebral/blob/master/pkg/controller/scale_manager.go#L208) should have high coverage.

High coverage on critical scale manager functionality.

#### Additional Context
N/A

ashleyschuett:
### Description

Using the https://github.com/kubernetes/sample-controller/blob/master/controller_test.go as a reference I have written tests to check that different scaling scenarios trigger the correct action. 

* Fixes #44 

 ### Testing

Tests are passing 

 #### Setup

`make test`

 #### Instructions

 ### Dependencies
N/A


pkg/controller/scale_manager_test.go
```diff
@@ -6,11 +6,367 @@ import (
...
+	f.kubeobjects = []runtime.Object{}
+	return f
+}
+
+var masterNodeTestLabels = map[string]string{
+	"node-role.kubernetes.io/master": "",
+}
+
+// TODO: how we do this, this feels wrong
+func NewStubEngine(name string) autoscaling.Engine {
```
ashleyschuett: I feel like putting this here is wrong, but I don't know where it should go :/

pkg/controller/scale_manager.go
```diff
@@ -51,7 +51,7 @@ func (d scaleDirection) String() string {
 type adjustmentType int
 
 const (
-	adjustmentTypeAbsolute = iota
+	adjustmentTypeAbsolute adjustmentType = iota
```
mattkelly: Nice.

pkg/controller/scale_manager_test.go
```diff
@@ -6,11 +6,367 @@ import (
...
+	nodeLister []*corev1.Node
+	// Objects from here preloaded into NewSimpleFake.
+	kubeobjects []runtime.Object
+	objects     []runtime.Object
+}
+
+func newFixture(t *testing.T) *fixture {
+	f := &fixture{}
+	f.t = t
+	f.objects = []runtime.Object{}
```
mattkelly: nit: `make([]runtime.Object)` would be consistent with our other code, though this is functionally equivalent.

pkg/controller/scale_manager_test.go
mattkelly: This isn't really a stub; it's more of a mock because `SetTargetNodeCount()` knows what to return based on what it's passed. Consider using a true mock by generating one using `mockery`. See the [prometheus backend tests](https://github.com/containership/cerebral/blob/master/pkg/metrics/backends/prometheus/prometheus_test.go#L112) for some examples of how to use a mock.

This is an ideal scenario to mock out the engine, since we're testing the scale manager which calls the engine and we would like to be able to determine if the engine was called or not (and potentially if it was called with the correct arguments).

pkg/controller/scale_manager_test.go
```diff
@@ -1,16 +1,389 @@
...
+
+// runHandleScaleRequestForASG lets us test handleScaleRequestForASG function to
+// get a deeper look into seeing that scale events are triggered as expected. That
+// means returning false when there was there was no scale event, and true when
+// one was triggered.
+func (f *fixture) runHandleScaleRequestForASG(asg *v1alpha1.AutoscalingGroup, req ScaleRequest, expectError bool, expectScale bool) {
+	c := f.newScaleManager()
+
+	scaled, err := c.handleScaleRequestForASG(asg, req)
+	if !expectError && err != nil {
```
mattkelly: Similar comment to above re: using `assert.NoError`, etc.

pkg/controller/scale_manager_test.go
```diff
@@ -1,16 +1,389 @@
...
+
+func TestNewScaleRequestWithNoASG(t *testing.T) {
+	f := newFixture(t)
+
+	req := newScaleRequest("asg-dne", scaleDirectionUp, adjustmentTypeAbsolute, true)
+
+	f.run(req)
+}
+
+func TestNewScaleRequesWithoutEngine(t *testing.T) {
```
mattkelly: Typo

pkg/controller/scale_manager_test.go
```diff
@@ -1,16 +1,389 @@
...
+
+	f.asgLister = append(f.asgLister, ag)
+	f.objects = append(f.objects, ag)
+
+	f.nodeLister = append(f.nodeLister, n)
+	f.kubeobjects = append(f.kubeobjects, n)
+	f.nodeLister = append(f.nodeLister, n2)
+	f.kubeobjects = append(f.kubeobjects, n2)
+
+	//mockEngine := mocks.Engine{}
```
mattkelly: Delete

pkg/controller/scale_manager_test.go
```diff
@@ -1,16 +1,389 @@
...
+	autoscaling.Registry().Put(engineName, &mocks.Engine{})
+	defer removeEngineFromAutoscalingRegistry(engineName)
```
mattkelly: nit: We're using a helper for registry `Delete()`, but not a helper for registry `Put()`. Any particular reason? I would be fine with not using any helpers, given that the `Delete()` helper is just one line 🤷‍♂️

pkg/controller/scale_manager.go
```diff
@@ -83,11 +82,8 @@ func adjustmentTypeFromString(s string) (adjustmentType, error) {
 type ScaleManager struct {
 	cerebralclientset cerebral.Interface
 
-	asgLister clisters.AutoscalingGroupLister
-	asgSynced cache.InformerSynced
-
+	asgLister  clisters.AutoscalingGroupLister
 	nodeLister corelistersv1.NodeLister
-	nodeSynced cache.InformerSynced
```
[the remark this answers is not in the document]
ashleyschuett: I was thinking the same thing but decided to remove it since this controller can't be called until an action happens from another controller, which all do wait for the cache sync. I'm fine with adding it back but didn't think it was needed.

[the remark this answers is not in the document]
mattkelly: This module should be decoupled from the knowledge of when it will be called, so we should add it. However, I'm also fine with adding a comment explaining this for now.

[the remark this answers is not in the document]
ashleyschuett: That is true, I will update it with the correct logic

pkg/controller/scale_manager_test.go
```diff
@@ -1,16 +1,390 @@
...
+	if !expectError {
+		assert.NoError(f.t, err)
+	}
+
+	if expectError {
+		assert.Error(f.t, err)
+	}
+
+	if err == nil && expectScale != scaled {
+		f.t.Errorf("expected scaled to be %t but instead was %t", expectScale, scaled)
```
mattkelly: Similar to a previous comment: is there a reason to not just use `assert.Equal(f.t, expectScale, scaled, "...")` here?

pkg/controller/scale_manager_test.go
```diff
@@ -1,16 +1,390 @@
...
+	}
+}
+
+func getKey(ag *v1alpha1.AutoscalingGroup, t *testing.T) string {
+	key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(ag)
+	if err != nil {
+		t.Errorf("Unexpected error getting key for foo %v: %v", ag.Name, err)
```
mattkelly: Just replace this conditional with `assert.NoError()`

PR #61, opened 2019-01-02
Add InfluxDB metrics backend

normanjoyner:
### Description

Adds InfluxDB as a valid metrics backend. Initial support includes `cpu_percent_utilization`, `memory_percent_utilization` and `custom` metrics.

* Fixes #56 

We may consider adding additional metric backend configuration options that can be passed when creating a new InfluxDB client such as username and password.

 ### Testing

 #### Setup
N/A

 #### Instructions
`make test`

Need to work on finalizing tests, so marking as WIP.

 ### Dependencies
N/A


mattkelly:
@normanjoyner rebase please

pkg/metrics/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,229 @@
...
+	config := metricConfiguration{}
+	if err := config.defaultAndValidate(configuration); err != nil {
+		return 0, errors.Wrap(err, "validating configuration")
+	}
+
+	switch metric {
+	case MetricCPUPercentUtilization.String():
+		query, err := buildCPUQuery(hostnames, configuration)
+		if err != nil {
+			return 0, errors.Wrap(err, "building query")
```
mattkelly: `building cpu query` would provide slightly more context (same for others in this `switch`)

pkg/metrics/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,229 @@
...
+
+func (b Backend) performQuery(db string, query string) (float64, error) {
+	log.Debugf("Performing InfluxDB query: %s", query)
+
+	res, err := b.influxDB.Query(influxdbclient.Query{
+		Command:  query,
+		Database: db,
+	})
+
+	if err != nil {
```
mattkelly: A `switch` might be easier on the eyes:

```
switch {
case err != nil:
	return 0, errors.Wrapf(err, "querying InfluxDB with string %q", query)

case res == nil:
	return 0, errors.Errorf("querying InfluxDB with string %q returned nil", query)

case res.Error() != nil:
	return 0, errors.Wrapf(res.Error(), "querying InfluxDB with string %q", query)

case len(res.Results) != 1:
	return 0, errors.New("querying InfluxDB return an unexpected number of results")
}
```

Up to you. I don't feel strongly about it after writing it out.

pkg/metrics/backends/influxdb/mocks/Client.go
```diff
@@ -0,0 +1,114 @@
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
mattkelly: (File comment): I forgot to mention in other PRs that added mocks that IMO we should rename the generated mock to be lowercase for consistency:

```
➜ fd mocks | xargs ls
mocks:

pkg/autoscaling/mocks:
Engine.go

pkg/autoscalingengine/mocks:
AutoscalingEngine.go

pkg/metrics/backends/influxdb/mocks:
Client.go

pkg/metrics/backends/prometheus/mocks:
prometheus_api.go
```

🤷‍♂️

pkg/metrics/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,229 @@
...
+	} else if res == nil {
+		return 0, errors.Errorf("querying InfluxDB with string %q returned nil", query)
+	} else if res.Error() != nil {
+		return 0, errors.Wrapf(res.Error(), "querying InfluxDB with string %q", query)
+	} else if len(res.Results) != 1 {
+		return 0, errors.New("querying InfluxDB return an unexpected number of results")
+	}
+
+	var result float64
+	switch v := interface{}(res.Results[0].Series).(type) {
```
mattkelly: This isn't needed. A `Series` is always a `Series` - it's a strict type: https://github.com/influxdata/influxdb/blob/master/client/v2/client.go#L503

Just remove the switch and do:

```
v := res.Results[0].Series

if len(v) != 1 ...

...

return result, nil
```

pkg/metrics/backends/influxdb/influxdb_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	emptyConfiguration = map[string]string{}
+
+	noHostnames       []string
+	oneHostname       = []string{"hostname-0"}
+	multipleHostnames = []string{"hostname-0", "hostname-1", "hostname-2"}
+)
+
+func TestNewClient(t *testing.T) {
+	// Should never fail with any valid URL because it's only constructing an
```
mattkelly: This was true for the Prometheus client but may not be true here. Based on the comment [here](https://github.com/containership/cerebral/pull/61/files#diff-fc678131f678e87540208cce8e26e067R48), it may be worth either changing this test case or adding another one or two? No need to go crazy of course because we're testing our layer, not theirs.

pkg/metrics/backends/influxdb/influxdb.go
ashleyschuett: 👍

pkg/metrics/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,227 @@
...
+		return 0, errors.Wrapf(res.Error(), "querying InfluxDB with string %q", query)
+
+	case len(res.Results) != 1:
+		return 0, errors.New("querying InfluxDB return an unexpected number of results")
+	}
+
+	var result float64
+	v := res.Results[0].Series
+	if len(v) != 1 || len(v[0].Values) != 1 {
+		return 0, errors.Errorf("expected Series to have a single value element but it has %d", len(v[0].Values))
```
mattkelly: If `len[v] == 0` then this line will explode when indexing into `v[0]`. I think this unfortunately needs to be split into two checks for proper error messaging.

pkg/metrics/backends/influxdb/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/pkg/errors"
+)
+
+// Metric is a metric exposed by this backend
+type Metric int
+
+const (
+	// MetricCPUPercentUtilization is used to gather info about the CPU usage of nodes
+	MetricCPUPercentUtilization Metric = iota
+	// MetricMemoryPercentUtilization is used to gather info about the CPU usage of nodes
```
mattkelly: Copy/paste error

pkg/metrics/backends/influxdb/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+const defaultDB = "telegraf"
+const defaultRange = "1m"
+const defaultRetentionPolicy = "rp_90d"
+
+// TODO consider splitting into multiple types instead of overloading this
+// single struct and ignoring irrelevant fields
+type metricConfiguration struct {
+	// -- Generic
+	Aggregation     string `json:"aggregation"`
+	DB              string `json:"db"`
```
mattkelly: Feels better to me to spell out `Database`, unless `DB` is a more typical "spelling" in the Influx ecosystem.

pkg/metrics/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,227 @@
...
+)
+
+// Backend implements a metrics backend for InfluxDB.
+type Backend struct {
+	influxDB influxdbclient.Client
+
+	nodeLister corelistersv1.NodeLister
+}
+
+// Average CPU usage across the given nodes for the given range
```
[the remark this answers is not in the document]
ashleyschuett: I don't understand what you mean here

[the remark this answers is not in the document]
mattkelly: The comment is saying that this takes the `average`, when in reality it will aggregate according to `Aggregation` (so it may take `max` or `min`, etc). Just a nit.

pkg/metrics/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,227 @@
...
+		return 0, errors.Wrapf(err, "querying InfluxDB with string %q", query)
+
+	case res == nil:
+		return 0, errors.Errorf("querying InfluxDB with string %q returned nil", query)
+
+	case res.Error() != nil:
+		return 0, errors.Wrapf(res.Error(), "querying InfluxDB with string %q", query)
+
+	case len(res.Results) != 1:
+		return 0, errors.New("querying InfluxDB return an unexpected number of results")
```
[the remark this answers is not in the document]
mattkelly: @ashleyschuett looks like you missed this one

pkg/metrics/backends/influxdb/influxdb.go
```diff
@@ -0,0 +1,231 @@
...
+	case res.Error() != nil:
+		return 0, errors.Wrapf(res.Error(), "querying InfluxDB with string %q", query)
+
+	case len(res.Results) != 1:
+		return 0, errors.New("querying InfluxDB returned an unexpected number of results")
+	}
+
+	var result float64
+	v := res.Results[0].Series
+	if len(v) == 0 {
```
mattkelly: Is `len(v) == 1` a better check?

ashleyschuett: do you mean `len(v) != 1` ?

mattkelly:
@ashleyschuett can you please update the main README in [relevant parts](https://github.com/containership/cerebral#metrics-backend) to reflect that we now support this metrics backend as well? There are multiple places (see the Project Status section as well - I think we should just rip out any references to specific backends there).

pkg/metrics/backends/influxdb/influxdb.go
mattkelly: Sure do 🤦‍♂️

ashleyschuett:
travis looks to have completed successfully, but isn't updating its status

Issue #52, opened 2018-12-20
Add DigitalOcean Kubernetes AutoscalingEngine


normanjoyner:
#### Why should this feature be added?
As a user of DIgitalOcean's managed Kubernetes service, I want to be able to autoscale my cluster. Currently, there is not a way to set up autoscaling natively on DigitalOcean, nor does the official cluster-autoscaler project have integration.

#### How would you like the feature to work?
The engine should integrate with the DigitalOcean API and support adding and removing node pool capacity.

#### What alternatives have been considered?
N/A

#### Additional Context
Kubernetes nodes launched through DigitalOcean's managed offering are not currently labeled with metadata such as cluster ID, node pool ID, etc. There may be a limitation to the initial version of this integration that requires users to hardcode the engine configuration to include a single node pool to autoscale.

mattkelly:
Good news regarding the limitations mentioned above! DO node pool label syncing is in-progress and should be available soon. See the following issues:

- https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/135
- https://github.com/digitalocean/digitalocean-cloud-controller-manager/issues/136

And the PR that's already up to implement it:

- https://github.com/digitalocean/digitalocean-cloud-controller-manager/pull/142

This should unblock us from having fully functional autoscaling through the DO Kubernetes offering. 🚀

PR #66, opened 2019-01-06
Add Digitalocean Autoscaling Engine

ashleyschuett:
### Description
Adds the Digitalocean Kubernetes offering as an engine that can be used with the autoscaler. 

* First iteration for #52 

 ### Testing
Build a docker image with the changes, and pushed to my docker hub. 
Added the deployment to run the cerebral instances on a kubernetes cluster provisioned through DO.  The following deployment assumes you have an admin service account with the proper permissions. 
```
apiVersion: extensions/v1beta1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/name: cerebral
    containership.io/app: cerebral
    containership.io/managed: "true"
  name: cerebral
  namespace: default
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/name: cerebral
      containership.io/app: cerebral
      containership.io/managed: "true"
  template:
    spec:
      containers:
      - env:
        - name: LOG_LEVEL
          value: DEBUG
        - name: DO_TOKEN
          valueFrom:
            secretKeyRef:
              key: DO_TOKEN
              name: do-env-secret
        image: ashleyschuett/cerebral
        imagePullPolicy: Always
        name: cerebral
      serviceAccount: admin
      tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/master
```
Created an ASE CR to add the Digitalocean engine to the running Cerebral deployment. 
```
apiVersion: cerebral.containership.io/v1alpha1
kind: AutoscalingEngine
metadata:
  name: digitalocean
spec:
  type: digitalocean
  configuration:
    clusterID: uuid
    nodePoolName: node-pool-name
    tokenEnvVarName: DO_TOKEN
```

Created a ASG that had `minNodes` of 4, my node pool of 3 was scaled up to 4. 

```
kind: AutoscalingGroup
metadata:
  name: do-asg
spec:
  cooldownPeriod: 600
  engine: digitalocean
  maxNodes: 4
  minNodes: 4
  suspended: false
```
 #### Setup

 #### Instructions

 ### Dependencies


docs/engines/digitalocean.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Description
+The DigitalOcean engine watches for scale events, which can be triggered if a node pool is not in bounds of the Autoscaling Group, or when threshold events are triggered from [Metrics Backends](/docs/metrics_backends).
+
+## Configuration
+In order for the DigitalOcean engine to scale your cluster, you will need to get the cluster ID and API token.
+You can find the cluster ID in the URL when looking at the cluster through the DigitalOcean dashboard.
+Finally, you can acquire an API token through the "Account" page on the DigitalOcean dashboard.
+
+There is also the option of adding `nodePoolLabelKey` which will allow for multiple Autoscaling Groups, one for each node pool, once DigitalOcean labels each node in a node pool with it's node pool information.
```
mattkelly: Grammar: `its`

pkg/autoscaling/engines/digitalocean/access_token.go
```diff
@@ -0,0 +1,15 @@
+package digitalocean
+
+import "golang.org/x/oauth2"
+
+type tokenSource struct {
+	AccessToken string
+}
+
+func (t *tokenSource) Token() (*oauth2.Token, error) {
```
mattkelly: A comment might be useful somewhere in this file indicating that we need to do this in order to implement some interface. I was confused why this additional indirection was required.

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+
+// getNodePoolByName uses the key assigned to 'NodePoolLabelKey' in the configuration
+// to get the DO node pool by ID
+func (e Engine) getNodePoolByLabel(nodeSelectors map[string]string) (*godo.KubernetesNodePool, error) {
+	poolID, ok := nodeSelectors[e.config.NodePoolLabelKey]
+	if !ok {
+		return nil, errors.New("node pool selector does not contain node pool key")
+	}
+
+	nodepool, _, err := e.client.Kubernetes.GetNodePool(context.TODO(), e.config.ClusterID, poolID)
```
mattkelly: Why are you using `context.TODO` here?

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+		// any scaling action
+		if total == numNodes {
+			return false, nil
+		}
+
+		if total != np.Count {
+			// calculate the number of nodes that should exist in the selected node pool.
+			// This only needs to be done if there are multiple node pools nodes being
+			// counted in the autoscaling group
+			numNodes = (numNodes - total) + np.Count
```
mattkelly: Can't this go below zero?

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		return e.scaleStrategyRandom(np, numNodes)
+	default:
```
mattkelly: nit: newline before `default` case.

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	log.Infof("DigitalOcean AutoscalingEngine %s is requesting DigitalOcean to scale to %d", e.Name(), numNodes)
+
+	switch strategy {
+	case "random", "":
+		// random is the default for this engine
+		np, total, err := e.getNodePool(nodeSelectors)
+		if err != nil {
+			return false, err
```
mattkelly: `errors.Wrap` is probably appropriate here

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+		return false, errors.Errorf("unable to scale node pool using strategy %s", strategy)
+	}
+}
+
+func (e Engine) getNodePool(nodeSelectors map[string]string) (*godo.KubernetesNodePool, int, error) {
+	total := 0
+	if len(nodeSelectors) > 0 {
+		np, err := e.getNodePoolByLabel(nodeSelectors)
+		if err != nil {
+			return nil, 0, err
```
mattkelly: `errors.Wrap` is probably appropriate here

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		return np, np.Count, nil
+	}
+
+	opts := godo.ListOptions{}
+	nodepools, _, err := e.client.Kubernetes.ListNodePools(context.TODO(), e.config.ClusterID, &opts)
+	if err != nil {
+		return nil, 0, err
```
mattkelly: `errors.Wrap` is probably appropriate here

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	for _, nodepool := range nodepools {
+		total += nodepool.Count
+	}
+
+	rand.Seed(time.Now().UnixNano())
```
mattkelly: `rand.Seed()` should not be called every time. It's not needed, plus it's slow, among other things. Simply call it once wherever it makes sense - probably an `init()` function?

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+
+	nodepool, _, err := e.client.Kubernetes.GetNodePool(context.TODO(), e.config.ClusterID, poolID)
+	if err != nil {
+		return nil, err
+	}
+
+	return nodepool, nil
+}
+
+// ScaleStrategyRandom takes in the number of desired nodes for a node pool.
```
mattkelly: nit: capital `S` here does not match function name. Maybe just omit the function name here for maintainability since it's not exported.

I also think this comment doesn't currently say much; shouldn't it mention that it picks a node randomly?

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+	// create a request to scale node pool
+	// both name and count are required fields
+	req := godo.KubernetesNodePoolUpdateRequest{
+		Name:  nodePool.Name,
+		Count: numNodes,
+	}
+
+	_, _, err := e.client.Kubernetes.UpdateNodePool(context.TODO(), e.config.ClusterID, nodePool.ID, &req)
+	if err != nil {
+		return false, errors.Wrap(err, "error scaling digitalocean nodepool")
```
mattkelly: `DigitalOcean node pool`

pkg/controller/autoscaling_engine.go
```diff
@@ -243,6 +244,16 @@ func instantiateEngine(engine *cerebralv1alpha1.AutoscalingEngine) (autoscaling.
 		}
 
 		return cae, nil
+	case "digitalocean":
```
mattkelly: nit: newline before

pkg/controller/autoscaling_engine.go
```diff
@@ -243,6 +244,16 @@ func instantiateEngine(engine *cerebralv1alpha1.AutoscalingEngine) (autoscaling.
...
+		// Ignore defensive checks on engine property values since validation happens
+		// upon new client creation. We're explicitly not copying the name and configuration
+		// here since it is assumed that NewClient will not modify the parameters
+		do, err := digitalocean.NewClient(engine.Name, engine.Spec.Configuration)
+		if err != nil {
+			return nil, errors.Wrapf(err, "constructing new digitalocean engine %q", engine.Name)
+		}
+
+		return do, nil
 	default:
```
mattkelly: nit: newline before

pkg/controller/autoscaling_engine.go
```diff
@@ -243,6 +244,16 @@ func instantiateEngine(engine *cerebralv1alpha1.AutoscalingEngine) (autoscaling.
 		}
 
 		return cae, nil
+	case "digitalocean":
+		// Ignore defensive checks on engine property values since validation happens
```
mattkelly: This comment is exactly the same as that for `containership`. Do we just need one comment?

pkg/autoscaling/engines/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,272 @@
...
+	"io/ioutil"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/digitalocean/godo"
+	"github.com/stretchr/testify/assert"
+)
+
+type fakeKubernetesService struct {
```
mattkelly: This feels like a lot of work for something that's not really a fake and could just be created using `mockery`. Am I misunderstanding?

pkg/autoscaling/engines/digitalocean/digitalocean.go
ashleyschuett: oh my understanding was that it wouldn't be random if only called once, and needed to be seeded for each call. Thank you for pointing that out, I will look into it more and fix that functionality

pkg/autoscaling/engines/digitalocean/digitalocean.go
ashleyschuett: yes, at which point the actual scale request will return a 500 and this will return an error. It will then retry the request and hopefully choose a different node pool. I feel like being any smarter than this and then the method is not random. I could see an argument for checking the `numNodes` for being 0 or less before making the request and returning an error. I'm just unsure of the correct thing to do if I check for 0 and they allow 0 later then I have to change the code, but it seems silly to allow 0 through now knowing its going to error if we check for < 0

docs/engines/digitalocean.md
```diff
@@ -0,0 +1,38 @@
+# DigitalOcean Engine
+
+## Limitations
+Cerebral Autoscaler works for DigitalOcean Kubernetes Clusters with one node pool, or multiple, but treats the whole cluster as one Autoscaling Group.
```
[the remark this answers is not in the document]
normanjoyner: i don't think saying `with one node pool, or multiple,` is required, it should be understood by default.

docs/engines/digitalocean.md
```diff
@@ -0,0 +1,38 @@
...
+In order for the DigitalOcean engine to scale your cluster, you will need to get the cluster ID and API token.
+You can find the cluster ID in the URL when looking at the cluster through the DigitalOcean dashboard.
+Finally, you can acquire an API token through the "Account" page on the DigitalOcean dashboard.
+
+There is also the option of adding `nodePoolLabelKey` which will allow for multiple Autoscaling Groups, one for each node pool, once DigitalOcean labels each node in a node pool with it's node pool information.  
+**Note:** The token that is used will need to have both read and write privileges to be able to scale the node pool.
+
+| Field | Required | Type | Description |
+| ----- | -------- | ---- | ----------- |
+| tokenEnvVarName | true | string | The environment variable name to use to get the DO API token |
```
normanjoyner: let's just spell out `DigitalOcean` as we've done elsewhere here

docs/engines/digitalocean.md
```diff
@@ -0,0 +1,38 @@
...
+| clusterID | true | string | The ID of the cluster that should be monitored and scaled |
+| nodePoolLabelKey | false | string | The label key that is used for labeling nodes as part of a node pool |
```
normanjoyner: i don't understand why this is an option

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,179 @@
...
+		// random is the default for this engine
+		np, total, err := e.getNodePool(nodeSelectors)
+		if err != nil {
+			return false, err
+		}
+
+		// The node pool count can be the total number of nodes in the cluster if a
+		// node selector is not supplied. Because of this we need to check the number
+		// of nodes in the selected node pool (or total for the cluster) before taking
+		// any scaling action
```
normanjoyner: the comment should explain why this is the case

pkg/autoscaling/engines/digitalocean/digitalocean.go
normanjoyner: in the scale down case we need to be prepared to make requests to multiple node pools to achieve the desired count.

docs/engines/digitalocean.md
ashleyschuett: So that when there is a standard node pool key set on DigitalOcean Kubernetes, this can be updated to reflect what that key is.

pkg/autoscaling/engines/digitalocean/digitalocean.go
ashleyschuett: I wasn't sure what context to use or if it makes sense to use the background context since the controller doesn't pass one to the engine.

pkg/autoscaling/engines/digitalocean/digitalocean.go
ashleyschuett: updated with this functionality

docs/engines/digitalocean.md
```diff
@@ -0,0 +1,38 @@
...
+## Description
+The DigitalOcean engine watches for scale events, which can be triggered if a node pool is not in bounds of the Autoscaling Group, or when threshold events are triggered from [Metrics Backends](/docs/metrics_backends).
+
+## Configuration
+In order for the DigitalOcean engine to scale your cluster, you will need to get the cluster ID and API token.
+You can find the cluster ID in the URL when looking at the cluster through the DigitalOcean dashboard.
+Finally, you can acquire an API token through the "Account" page on the DigitalOcean dashboard.
+
+There is also the option of adding `nodePoolLabelKey` which will allow for multiple Autoscaling Groups, one for each node pool, once DigitalOcean labels each node in a node pool with its node pool information.  
+**Note:** The token that is used will need to have both read and write privileges to be able to scale the node pool.
```
mattkelly: Doesn't this belong after the preceding paragraph, not here?

docs/engines/digitalocean.md
mattkelly: I'm also confused by this. Can't the user just update the `nodeSelector` in the ASG CR(s)? Isn't this actually less flexible?

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+	"context"
+	"os"
+
+	"github.com/pkg/errors"
+	"golang.org/x/oauth2"
+
+	"github.com/containership/cerebral/pkg/autoscaling"
+	"github.com/containership/cluster-manager/pkg/log"
+
+	"github.com/digitalocean/godo"
```
mattkelly: Can we start enforcing the convention of having `containership` internal packages last?

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+		return nil, err
+	}
+
+	token := os.Getenv(configuration["tokenEnvVarName"])
+	oauthClient := oauth2.NewClient(oauth2.NoContext, &tokenSource{
+		AccessToken: token,
+	})
+	doClient, err := godo.New(oauthClient)
+	if err != nil {
+		return nil, errors.Errorf("failed to create godo client: %s", err)
```
mattkelly: Why not just:

`errors.Wrap(err, "creating godo client")`

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+			ClusterID:        configuration["clusterID"],
+			NodePoolLabelKey: configuration["nodePoolLabelKey"],
+		},
+		client: doClient,
+	}
+
+	return e, nil
+}
+
+func validateConfiguration(configuration map[string]string) error {
```
mattkelly: I think the de facto convention in golang is to have private functions below exported functions ("most important to least important").

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+		return false, errors.New("cannot scale below 0")
+	}
+
+	log.Infof("DigitalOcean AutoscalingEngine %s is requesting DigitalOcean to scale to %d", e.Name(), numNodes)
+
+	switch strategy {
+	// random is the default for this engine
+	case "random", "":
+
+		var scaled bool
```
mattkelly: nit: no blank line before this line

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+
+		if err != nil {
+			return false, errors.Wrap(err, "unable to scale DigitalOcean cluster")
+		}
+
+		return scaled, nil
+
+	default:
+		return false, errors.Errorf("unable to scale node pool using strategy %s", strategy)
```
mattkelly: Something more like `unknown scale strategy` feels more relevant to me

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+		return false, err
+	}
+
+	if np.Count == numNodes {
+		return false, nil
+	}
+
+	err = e.scaleNodePoolByCount(np, numNodes)
+	if err != nil {
+		return false, errors.Wrapf(err, "unable to scale node pool with label %s", nodeSelectors)
```
mattkelly: `with node selectors`? Or at least plural `labels`

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	return true, nil
+}
+
+// Node pools currently do not have labels we need to scale randomly.
```
mattkelly: Reads weird

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+
+	var total int
+	for _, np := range nps {
+		total += np.Count
+	}
+
+	switch {
+	case total < numNodes:
+		scaleUpBy := numNodes - total
+		err = e.scaleNodePoolByCount(nps[0], scaleUpBy)
```
mattkelly: Probably worth noting that this isn't actually random. Or maybe just make it random? That feels like the better bet to me.

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if numToScale != 0 {
+		return errors.New("unable to scale to desired node count")
+	}
+
+	return nil
+}
+
+func getNodepoolCount(curr, total int) int {
```
mattkelly: `getNodePoolCount`

This could probably use a better name or a comment

pkg/autoscaling/engines/digitalocean/digitalocean.go
mattkelly: I think using `context.Background()` makes sense. Though they are exactly the same in practice, it feels unlikely to me that we would ever extend this code to use a passed-in context (especially if the scale request ends up coming over gRPC).

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+			NodePoolLabelKey: configuration["nodePoolLabelKey"],
+		},
+		client: doClient,
+	}
+
+	return e, nil
+}
+
+func validateConfiguration(configuration map[string]string) error {
+	if configuration["clusterID"] == "" {
```
mattkelly: The config keys are used in a few places; probably worth pulling out into constants.

```
const (
	ConfigKeyClusterID ...
	ConfigKeyTokenEnvVarName ...
    ...
)
```

🤷‍♂️

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	if err != nil {
+		return false, errors.Wrap(err, "unable to scale random node pool")
+	}
+
+	return true, nil
+}
+
+func (e Engine) randomScaleDown(nodepools []*godo.KubernetesNodePool, numToScale int) error {
```
mattkelly: Worth noting this also isn't random.

pkg/autoscaling/engines/digitalocean/digitalocean.go
```diff
@@ -0,0 +1,245 @@
...
+		err := e.scaleNodePoolByCount(np, scaleNodePoolTo)
+		if err != nil {
+			return err
+		}
+
+		numToScale = numToScale - (np.Count - scaleNodePoolTo)
+	}
+
+	if numToScale != 0 {
+		return errors.New("unable to scale to desired node count")
```
mattkelly: Probably worth noting why, or at least documenting why this could happen

pkg/autoscaling/engines/digitalocean/digitalocean_test.go
```diff
@@ -0,0 +1,228 @@
...
+				ID:   "node-1",
+				Name: "test-droplet",
+			},
+		},
+	}
+}
+
+func newMockClient() *godo.Client {
+	client := godo.NewClient(nil)
+	//client.Kubernetes = &mocks.KubernetesService{}
```
mattkelly: Dead code

pkg/autoscaling/engines/digitalocean/mocks/KubernetesService.go
```diff
@@ -0,0 +1,401 @@
+// Code generated by mockery v1.0.0. DO NOT EDIT.
```
mattkelly: (File comment): We're trying to [move forward with renaming the generated mocks to match our file naming conventions](https://github.com/containership/cerebral/pull/61#discussion_r245377471).

pkg/controller/autoscaling_engine.go
```diff
@@ -229,7 +230,7 @@ func (c *AutoscalingEngineController) syncHandler(key string) error {
 	return nil
 }
 
-// insantiateEngine instantiates a new engine for the given AutoscalingEngine.
+// instantiateEngine instantiates a new engine for the given AutoscalingEngine.
```
mattkelly: 👌

docs/engines/digitalocean.md
ashleyschuett: Yes, but out of the node selector, we need to know how to map that to the id of the node pool to scale. this is the same thing we are doing in the containership engine

pkg/autoscaling/engines/digitalocean/digitalocean.go
ashleyschuett: 👍

docs/engines/digitalocean.md
mattkelly: Ahhh right, makes sense. Thanks.

PR #90, opened 2019-02-05
Refactor node selection in autoscaling engines 

### LINKED ISSUE #71: Address inflexibility related to AutoscalingGroup nodeSelector
When a user defines an `AutoscalingGroup`, they do so by specifying a `nodeSelector` to target a subset of cluster nodes to collect metrics against. Since the `nodeSelector` is an arbitrary key/value label pair on a node, a user is able to create `AutoscalingGroup`s that span logical "node pools", as implemented by the engine (e.g. Containership, DigitalOcean, etc).

This is flexible in that it allows users to arbitrarily remap the concept of a "node pool". For example, a user could make a single `AutoscalingGroup` out of an entire Kubernetes cluster, regardless of the number of underlying "node pools". Doing so reinforces Cerebral concepts such as a `scalingStrategy` in that Cerebral's engine implementations could be intelligent enough to make decisions such as:
* scale up the pool with the least expensive nodes
* scale down the pool with the most expensive nodes
* scale up the pool with the "largest" instance type
* ... and so on

However, in the current implementation, we expect the `nodeSelector` as defined by the user, to match that which is defined by the engine implementation, and hold the pool ID. 

The Containership engine is a concrete example of this:
https://github.com/containership/cerebral/blob/3b93cf7ba12040fda0ac7285cb416eba1ec11723/pkg/autoscaling/engines/containership/containership.go#L94

Additionally, the DigitalOcean engine mimics this limitation by assuming if a `nodeSelector` is supplied on the `AutoscalingGroup`, the end user must also specify the same selector on the engine, and it still must point to a label holding the node pool ID:
https://github.com/containership/cerebral/blob/3b93cf7ba12040fda0ac7285cb416eba1ec11723/pkg/autoscaling/engines/digitalocean/digitalocean.go#L247

If an engine must be opinionated about how `AutoscalingGroup`s are defined, it should be explicitly stated that an `AutoscalingGroup` is a one-to-one map to the concept of a "node pool". In this model, it no longer makes sense for the user to define the composition of an `AutoscalingGroup` through a `nodeSelector`, and we should use a concept similar to a `topologyKey`.

However, since the entire `nodeSelector` is passed through to the engine from the scale manager, I would expect we can first reselect the related nodes, and expect a well-defined label to exist on all of those nodes (e.g. `containership.io/node-pool-id` in the case of the Containership engine).
https://github.com/containership/cerebral/blob/cbb40c31a889d5485a7a4a167c5eeb69883652ce/pkg/controller/scale_manager.go#L258

#### Additional Context
N/A

ashleyschuett:
### Description

This refactors the Containership and DigitalOcean engines to use the node selectors to reselect the nodes in the cluster. It will then choose a random node from that selection and grab the node pool ID of it using the known label key. 

I have also ripped out the `buildNodeLister` function from multiple tests and am instead referencing the new `testdata` package that contains the function. This package gets ignored when building godocs which is why I used it. However, I'm not sure it's entirely correct as it seems I should only store data in the package and not common functions, but I have yet to figure out something better. 

* Fixes #71 

This makes scaling back up once the selected node pool goes to 0 impossible. Since if zero nodes match the node selector then we have nothing to get the node pool id from using the label key. 
^^ This has been made into its' own ticket #92 to be discussed, it will be considered separately from this PR. 
 ### Testing

 #### Setup

 #### Instructions

 ### Dependencies


pkg/autoscaling/engines/containership/containership.go
```diff
@@ -71,16 +78,16 @@ func (e Engine) SetTargetNodeCount(nodeSelectors map[string]string, numNodes int
...
-	}
-
 	log.Infof("Containership AutoscalingEngine %s is requesting Containership Cloud to set target nodes %v to %d", e.Name(), nodeSelectors, numNodes)
 
 	switch strategy {
 	case "random", "":
 		// random is the default for this engine
+		id, err := engineutils.GetNodePoolIDToScale(nodeSelectors, nodePoolIDLabelKey, numNodes, e.nodeLister)
+		if err != nil {
+			return false, errors.Wrap(err, "Containership enigne unable to scale")
```
mattkelly: Spelling

mattkelly: In general I think these messages could be more specific and consistent with how we're writing `Wrap{,f}` messages elsewhere. For example: `errors.Wrap(err, "getting node pool ID to scale")`

pkg/autoscaling/engines/utils/nodes.go
```diff
@@ -0,0 +1,53 @@
...
+
+	ns := nodeutil.GetNodesLabelSelector(selector)
+
+	return nodeLister.List(ns)
+}
+
+// GetNodePoolIDToScale selects all the nodes that match the passed in node selector
+// it then checks for any errors that could have happened selecting nodes
+// and finally returns the ID for the node pool that should be scaled
+func GetNodePoolIDToScale(nodeSelectors map[string]string, knownLabelKey string, numNodes int, nodeLister corelistersv1.NodeLister) (string, error) {
```
mattkelly: Consider splitting into multiple lines

pkg/autoscaling/engines/utils/nodes.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"math/rand"
+
+	corev1 "k8s.io/api/core/v1"
+
+	corelistersv1 "k8s.io/client-go/listers/core/v1"
+
+	"github.com/containership/cerebral/pkg/nodeutil"
+	"github.com/pkg/errors"
```
mattkelly: This belongs higher up - convention elsewhere has been to put non-k8s imports immediately after the standard library imports (i.e. after `math/rand` here).

pkg/autoscaling/engines/utils/nodes.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+// GetASGNodes returns all nodes that match the passed in node selector
+func GetASGNodes(selector map[string]string, nodeLister corelistersv1.NodeLister) ([]*corev1.Node, error) {
+	if nodeLister == nil {
+		return nil, errors.New("node lister can not be nil")
```
mattkelly: nit: `cannot` consistency

pkg/autoscaling/engines/utils/nodes.go
```diff
@@ -0,0 +1,57 @@
...
+		return "", errors.Wrap(err, "unable to list nodes")
+	}
+
+	numSelectedNodes := len(nodes)
+	if numSelectedNodes == 0 {
+		return "", errors.New("no nodes selected")
+	}
+
+	if numSelectedNodes == numNodes {
+		return "", errors.New("can not scale to current count")
```
mattkelly: nit: `cannot` consistency

pkg/autoscaling/engines/utils/nodes.go
```diff
@@ -0,0 +1,57 @@
...
+
+	ns := nodeutil.GetNodesLabelSelector(selector)
+
+	return nodeLister.List(ns)
+}
+
+// GetNodePoolIDToScale selects all the nodes that match the passed in node selector
+// it then checks for any errors that could have happened selecting nodes
+// and finally returns the ID for the node pool that should be scaled
+func GetNodePoolIDToScale(
```
mattkelly: This isn't the right utility function. This is specifically for picking something using the `random` strategy. I don't think this should be a shared function - IMO it's fine for each engine to have to implement it.

pkg/autoscaling/engines/utils/nodes.go
mattkelly: Is this really an error? I would argue that we should just return `false, nil` from the engine in this case (it may also be engine-dependent). I think this is what we were doing before?

pkg/testdata/utils.go
```diff
@@ -0,0 +1,32 @@
+package testdata
```
mattkelly: Is `testdata` the right package name? I would argue it's not because it's reserved for [test fixtures](https://dave.cheney.net/2016/05/10/test-fixtures-in-go) which are typically just data files. We could name this using the `*test` package naming convention similar to [`httptest`](https://golang.org/pkg/net/http/httptest/). I'm not sure of a great name though since this is useful for both metrics backends and engines.

mattkelly:
@ashleyschuett as discussed, I think it makes sense to get this in and then worry about #92

pkg/autoscaling/engines/utils/nodes.go
ashleyschuett: I thought about this as well, but returning an empty id with no error doesn't feel quite right since the function is called `GetNodePoolIDToScale` I would expect an id or an error. Having no error and no id feels wrong to me. 

I could comment it to make it more clear? Or if there's an argument for having `return "", nil` be valid? I do agree it could be engine-dependent which is why I don't like either solution.

mattkelly: IMO we need to do whatever needs to happen in order to return `false, nil` at the engine interface level. I don't think returning an error makes sense. The engine is simply deciding to take no action.

ashleyschuett: guess your next comment makes this not matter. 👋

pkg/testdata/utils.go
ashleyschuett: yeah, that's what I was trying to express in the description. I don't like making it its own package, and its really just a helper function and not creating a testing object but... what about making a `kubernetestest` package and we can start pulling other kubernetes testing things that make sense in there?

mattkelly: `kubernetestest` sounds like a decent name for this package for now

pkg/autoscaling/engines/utils/nodes.go
ashleyschuett: is it worth it to have this package at all then? the only other function probably doesn't need its own package. 🤔

mattkelly: Yeah I was also previously going back-and-forth on whether or not the above function is useful enough to pull out. I don't think I have a strong opinion.

mattkelly:
@ashleyschuett rebase please

Issue #69, opened 2019-01-22
Add gpu_percent_allocation metric to Kubernetes metrics backend


normanjoyner:
#### Why should this feature be added?
As a user, I want to be able to scale my AutoscalingGroup up and down based on GPU allocation, similar to how I can with cpu, memory, storage, and pods.
#### How would you like the feature to work?
This metric should calculate the total GPU allocation percentage across an AutoscalingGroup. 

#### What alternatives have been considered?
N/A

#### Additional Context
N/A

PR #91, opened 2019-02-05
Allocatable GPU resource support

nicktate:
### Description

* added amd and nvidia gpu vendor support to kubernetes metrics backend

* Fixes https://github.com/containership/cerebral/issues/69

* We need to add support scaling from 0, as that will be a common case especially for GPU types since they are an extremely expensive instance

 ### Testing

* Added in new set of unit tests for GPU resources. Checks both amd and nvidia vendors (currently the only two supported on kubernetes)

 #### Instructions

Create policy to scale the `gpu_percent_allocation` metric


pkg/metrics/backends/kubernetes/kubernetes.go
```diff
@@ -115,6 +118,34 @@ func (b Backend) calculateCPUAllocationPercentage(pods []*corev1.Pod, nodes []*c
 	return (100 * (float64(requestedCPUs) / float64(allocatableCPUs)))
 }
 
+func (b Backend) calculateGPUAllocationPercentage(pods []*corev1.Pod, nodes []*corev1.Node) float64 {
+	log.Infof("Performing gpu allocation calculation of %d pods across %d nodes", len(pods), len(nodes))
```
mattkelly: `Debugf` please

pkg/metrics/backends/kubernetes/kubernetes_test.go
```diff
@@ -1,10 +1,11 @@
 package kubernetes
 
 import (
-	"github.com/stretchr/testify/assert"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
```
mattkelly: 👍

pkg/metrics/backends/kubernetes/kubernetes_test.go
```diff
@@ -149,6 +157,24 @@ func TestCalculateCPUAllocationPercentage(t *testing.T) {
 	assert.Equal(t, float64(10), percentage, "returns correct allocation percentage")
 }
 
+func TestCalculateGPUAllocationPercentage(t *testing.T) {
+	// 6 total gpus, 3 nvidia requested
+	var podList = []*corev1.Pod{pod0}
+	var nodeList = []*corev1.Node{node0, node1, node2}
+	percentage := backend.calculateGPUAllocationPercentage(podList, nodeList)
+	assert.Equal(t, float64(50), percentage, "returns correct allocation percentage")
```
mattkelly: nit: Maybe augment or replace this message with the info in the comments so they're not all the same?

pkg/metrics/backends/kubernetes/metrics.go
```diff
@@ -14,11 +16,21 @@ const (
 	MetricPodPercentAllocation
 )
 
+// GPUVendors returns array of supported GPU vendors
+func GPUVendors() [2]string {
```
mattkelly: Any particular reason this is a function rather than a static array? Maybe because arrays can't be `const`? I think most/all of the things in this file can/should actually be unexported which might make us feel better about having these in a simple array.

I don't really have a strong opinion and am fine with leaving this as-is. Mostly just thinking aloud.

pkg/metrics/backends/kubernetes/kubernetes.go
```diff
@@ -2,7 +2,6 @@ package kubernetes
 
 import (
 	"github.com/pkg/errors"
-
```
mattkelly: Add this back

pkg/metrics/backends/kubernetes/metrics.go
nicktate: Yeah, I was doing a little research into golang, it does not look like they have support for array constants so that's why I opted to go this route. If you have alternative suggestions, I'm definitely open to a better way to structure it.

nicktate:
updated @mattkelly

pkg/metrics/backends/kubernetes/metrics.go
```diff
@@ -14,11 +16,21 @@ const (
...
+	return [...]string{
```
ashleyschuett: just curious why you chose this syntax?

nicktate: `...` refers to whatever the length of an array is when defined as an array literal. You unfortunately cannot define that at the function return level because it can only be referenced explicitly with a literal, so I just opted to return a fixed length of the size of the return value.

PR #97, opened 2019-02-19
Update Cerebral to use client-go v10.0.0

ashleyschuett:
### Description

Update packages to support Kubernetes 1.13, with this update the code gen also needed to be updated as there were changes to the method call for a Patch() request.  

 ### Testing

`make test`

 #### Setup

 #### Instructions

 ### Dependencies


Gopkg.toml
```diff
@@ -56,7 +55,23 @@ required = [
...
 
+[[constraint]]
+  name = "k8s.io/code-generator"
+  version = "kubernetes-1.13.3"
+
 [prune]
   go-tests = true
   # Note that we can't do this due to the code generator packages required; see above
-  #unused-packages = true
+  unused-packages = true
```
mattkelly: Nice 😎

pkg/client/informers/externalversions/factory.go
```diff
@@ -32,36 +32,79 @@ import (
...
+// WithTweakListOptions sets a custom filter on all listers of the configured SharedInformerFactory.
+func WithTweakListOptions(tweakListOptions internalinterfaces.TweakListOptionsFunc) SharedInformerOption {
+	return func(factory *sharedInformerFactory) *sharedInformerFactory {
+		factory.tweakListOptions = tweakListOptions
+		return factory
+	}
+}
+
+// WithNamespace limits the SharedInformerFactory to the specified namespace.
+func WithNamespace(namespace string) SharedInformerOption {
```
mattkelly: Some useful new functions here which I think you've mentioned before. We can probably use them over in cluster-manager at least.

PR #104, opened 2019-03-13
clean up poll manager channels

ashleyschuett:
### Description

This PR fixes channels being left open after a goroutine process returns. 

We are now always closing the `stopCh` that is created for each poll manager, as well as closing the `alertCh` that is created for each poll manager, and the `errCh`. 

 ### Testing

- Created an image with pprof tools and this change. 

- the original image goroutine this is after running for a day, however both routines increased by ~100 in the hour I was testing. 
```
goroutine profile: total 2260
1946 @ 0x42db5b 0x42dc03 0x404ead 0x404c85 0x1291bc6 0x12916d6 0x45b071
#    0x1291bc5    github.com/containership/cerebral/pkg/controller.(*metricPoller).fireAlert+0xb5    /go/src/github.com/containership/cerebral/pkg/controller/metric_poller.go:101
#    0x12916d5    github.com/containership/cerebral/pkg/controller.metricPoller.run+0x485        /go/src/github.com/containership/cerebral/pkg/controller/metric_poller.go:88

226 @ 0x42db5b 0x42dc03 0x404ead 0x404c85 0x1291bc6 0x1291714 0x45b071
#    0x1291bc5    github.com/containership/cerebral/pkg/controller.(*metricPoller).fireAlert+0xb5    /go/src/github.com/containership/cerebral/pkg/controller/metric_poller.go:101
#    0x1291713    github.com/containership/cerebral/pkg/controller.metricPoller.run+0x4c3        /go/src/github.com/containership/cerebral/pkg/controller/metric_poller.go:82
```

- new image, this has been holding the same number of goroutines in the hour it has been running
```
goroutine profile: total 85
```

 #### Setup

 #### Instructions

 ### Dependencies


pkg/controller/poll_manager.go
```diff
@@ -98,17 +100,23 @@ func (m pollManager) run() error {
 
 			err := <-errCh
 			if err != nil {
-				// If a scale request fails, just return an error so the relevant ASG can be re-enqueued
-				return errors.Wrap(err, "requesting scale manager to scale")
+				// If a scale request fails, save the error to the result to return and close the stop channel
+				// The error needs to be saved and returned so the relevant ASG can be re-enqueued
+				result = errors.Wrap(err, "requesting scale manager to scale")
```
mattkelly: IMO, it would be more clear if we closed the stopCh [here](https://github.com/mattkelly/cerebral/blob/1631e6ab560a4d914652973739f154e1329e279c/pkg/controller/metrics.go#L365-L370), so that this module is managing only individual metric pollers, while the metrics controller manages the poll managers themselves. It just feels like a better separation of concerns to me. It took me a few reads to grok what was going on here. Thoughts?

ashleyschuett: Yeah that makes sense to be I can change it, then I can also call the`cleanupPollManagerForASG` and we will only be closing the `stopCh` in one spot.

ashleyschuett: oh wait, now I remember why I did not do that. The poller manager will no longer be listening on stopCh if we return since that `for` loop won't be running anymore.

mattkelly: Ahh I see. This seems like a potentially poor design on my part 🤔. Let's go with this solution then.

mattkelly:
Excellent work! Thanks for hunting this down. 🥇

PR #105, opened 2019-03-15
Fix/cleanup pollers

ashleyschuett:
### Description

This cleans up the logic for stopping channels. 

We realized that the metric_poller was blocking an ASG restart if the `stopCh` closed when the poller was in ` case <-ticker.C` when the poll manager returned. This was fixed by adding `default` case to the poll manager so that the `alertCh` does not block if the alert is unable to be sent.

Thank you @mattkelly for the assist. 🎈

The fix made for PR #104 exposed an edge case that added a bug. The cerebral pod was periodically restarting due to trying to close an already closed channel. 

```
NAME                                        READY   STATUS    RESTARTS   AGE
cerebral-76447967cc-nfq67                   1/1     Running   5          19h

panic: close of closed channel
goroutine 139 [running]:
github.com/containership/cerebral/pkg/controller.pollManager.run(0xc00057db90, 0x24, 0xc0006f9ef0, 0xc0009f4060, 0x1817380, 0xc0003336c0, 0xc000084180, 0xc0000fb980, 0x7974222020202020, 0x626f22203a226570)
	/go/src/github.com/containership/cerebral/pkg/controller/poll_manager.go:106 +0x5b1
github.com/containership/cerebral/pkg/controller.(*MetricsController).syncHandler.func1(0xc00057db90, 0x24, 0xc000480000, 0xc000630000)
	/go/src/github.com/containership/cerebral/pkg/controller/metrics.go:365 +0x12a
created by github.com/containership/cerebral/pkg/controller.(*MetricsController).syncHandler
	/go/src/github.com/containership/cerebral/pkg/controller/metrics.go:362 +0x635
root@70fac0b5-7717-4e83-86c0-43728d1fed5e:~# packet_write_wait: Connection to 104.248.124.73 port 22: Broken pipe
```

 ### Testing

- Built pod, and made sure there was no goroutine leak, which is staying at a constant number and not growing 

```
goroutine profile: total 88
```

- I added and deleted ASP to trigger different cases and made sure the pod was not restarting and gracefully handling the changes. 

 #### Setup

 #### Instructions

 ### Dependencies

### DIFF
--- a/pkg/controller/metric_poller.go
+++ b/pkg/controller/metric_poller.go
@@ -41,6 +41,14 @@ func newMetricPoller(asp *v1alpha1.AutoscalingPolicy, nodeSelector map[string]st
 	}
 }
 
+func sendAlert(alertCh chan<- alert, msg alert) {
+	select {
+	case alertCh <- msg:
+	default:
+		log.Debugf("Alert channel is full. Discarding alert %v", msg)
+	}
+}
+
 func (p metricPoller) run(wg *sync.WaitGroup, alertCh chan<- alert, stopCh <-chan struct{}) {
 	defer wg.Done()
 
@@ -63,14 +71,14 @@ func (p metricPoller) run(wg *sync.WaitGroup, alertCh chan<- alert, stopCh <-cha
 			backend, err := metrics.Registry().Get(backendName)
 			if err != nil {
 				err = errors.Wrapf(err, "metrics backend %q specified by policy %q is unavailable", backendName, policyName)
-				alertCh <- alert{err: err}
+				sendAlert(alertCh, alert{err: err})
 				return
 			}
 
 			val, err := backend.GetValue(metric, metricConfig, p.nodeSelector)
 			if err != nil {
 				err = errors.Wrapf(err, "getting metric %q for policy %q", metric, policyName)
-				alertCh <- alert{err: err}
+				sendAlert(alertCh, alert{err: err})
 				return
 			}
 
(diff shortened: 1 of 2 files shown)


pkg/controller/poll_manager.go
```diff
@@ -100,23 +119,15 @@ func (m pollManager) run() error {
...
-			close(alertCh)
-			close(errCh)
-
-			log.Infof("Poll manager for AutoscalingGroup %s shutdown success", m.asgName)
-			return result
+			log.Infof("Poll manager for AutoscalingGroup %s shutting down", m.asgName)
+			return nil
+		// we don't want to block on the alertCh, so if the channel can't keep up
+		// we should drop the alert
+		default:
```
mattkelly: This is not quite right. The `default` case should be [on the producer side](https://stackoverflow.com/a/25657232), so we'll need a little helper function over there for any sends. I don't think this does anything as-is. We should also add a debug log in that default case to see when messages are dropped.

Note that a ring buffer (ring channel) might also be appropriate here. In that case, the oldest message would be dropped instead of the newest. However, I think just dropping the newest one is fine for our case (though admittedly I haven't thought about the implications too much).

pkg/controller/metric_poller.go
```diff
@@ -41,6 +41,14 @@ func newMetricPoller(asp *v1alpha1.AutoscalingPolicy, nodeSelector map[string]st
 	}
 }
 
+func safelySendAlert(alertCh chan<- alert, msg alert) {
+	select {
+	case alertCh <- msg:
+	default:
+		log.Infof("Alert channel is full. Discarding alert %v", msg)
```
mattkelly: `log.Debugf`

pkg/controller/metric_poller.go
```diff
@@ -41,6 +41,14 @@ func newMetricPoller(asp *v1alpha1.AutoscalingPolicy, nodeSelector map[string]st
 	}
 }
 
+func safelySendAlert(alertCh chan<- alert, msg alert) {
```
mattkelly: This function doesn't strike me as adding any safety per se (maybe if it checked that the channel wasn't closed first, but we know that won't happen). Maybe just `sendAlert`?

PR #109, opened 2019-04-08
add known key to digitalocean engine

### LINKED ISSUE #87: Refactor DigitalOcean Autoscaling Engine to use node pool information
### Description
During the initial implementation of the DigitalOcean autoscaling engine, DigitalOcean did not automatically label nodes with their node pool information. Because of this, the implementation treats the whole cluster as one node pool no matter how many node pools are being used.

#### How would you like the feature to work?
Once the node pool information is added to each node we should refactor the engine to use the node pool label key to know which node pool should be interacted with during scaling events. 

In the [previous ticket](https://github.com/containership/cerebral/issues/52) to add the engine we were tracking the progress of this issue. We can still use as it lists the current issues and discussion the DigitalOcean is having about adding these labels.

ashleyschuett:
### Description

Remove the configuration option for the user to choose their own well known key for how to divide nodes in to groups. 

* Fixes #87 

This change gets rid of a lot of DigitalOceans' engines flexibility, but also removes the burden of groups nodes of the user. This change makes the DO engine now mirror the Containership engine. Is the removal of flexibility for node groupings desired?  

 ### Testing

- created docker image and pushed to my docker registry
- ran new image on DOKS cluster
- increased CPU requests to >70% and after sample period the cluster scaled up the node pool by desired count. 

 #### Setup

 #### Instructions

 ### Dependencies

### DIFF
--- a/pkg/autoscaling/engines/digitalocean/digitalocean_test.go
+++ b/pkg/autoscaling/engines/digitalocean/digitalocean_test.go
@@ -162,3 +162,33 @@ func TestSetTargetNodeCount(t *testing.T) {
 	assert.NoError(t, err)
 	assert.False(t, result)
 }
+
+func TestSetTargetNodeCountDoResponseError(t *testing.T) {
+	nodeLister := kubernetestest.BuildNodeLister([]corev1.Node{node0})
+	c, kmocks := fakeAutoscalingEngine(nodeLister)
+
+	nodepool := newFakeNodePool(nodePoolID, nodePoolName, 1)
+	nodepools := []*godo.KubernetesNodePool{nodepool}
+
+	label := map[string]string{
+		nodePoolIDLabelKey: nodePoolID,
+	}
+
+	kmocks.On("GetNodePool", mock.Anything, mock.Anything, mock.Anything).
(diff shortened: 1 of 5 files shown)


mattkelly:
As discussed in Slack, this change makes sense to me. I don't think the flexibility gained by allowing a user to specify node label selectors is worth the complexity and probable confusion in cases where we have well-known labels available to us. Placing the onus on the user in well-defined cases like this doesn't seem very friendly.

pkg/autoscaling/engines/digitalocean/digitalocean_test.go
```diff
@@ -190,83 +157,8 @@ func TestSetTargetNodeCount(t *testing.T) {
 	kmocks.On("GetNodePool", mock.Anything, mock.Anything, mock.Anything).
 		Return(nil, nil, errors.New("node pool not found")).Once()
 
-	label[key] = "bad-id"
+	label[nodePoolIDLabelKey] = "id-does-not-match-any-nodes"
```
mattkelly: 👍

ashleyschuett:
updated with testing information, also added tests to check for transient DO errors to increase testing coverage

PR #111, opened 2019-04-17
Update getNodeExporterNodes to use Prometheus

### LINKED ISSUE #15: Don't hardcode prom-exporter=node label selector for prometheus backend
See these discussions:

- https://github.com/containership/cerebral/pull/13#discussion_r238404175
- https://github.com/containership/cerebral/pull/13#discussion_r238404797

We shouldn't hardcode the label selector as it's specific to our setup. Let's see if we can refactor that block of code to use a dynamic label selector built from the Prometheus `targets` response (can be retrieved using `prometheusclient.Targets()`. Or maybe we can somehow grab the `PodIP`s we need directly from the targets.

#### Additional Context

Pulling this into the release so we can try to get it in, but I don't think it should block the release.

noah-vienneau:
### Description

Eliminates the need to use a label selector in favor of the prometheus Targets api call, which gets rid of needing to hard code "prom-exporter". All changes made to the getNodeExporterPodIPsOnNodes function.  
* Fixes #15 

 ### Testing
No changes to the existing tests
Only tested using Unit Tests

### DIFF
--- a/pkg/metrics/backends/prometheus/prometheus.go
+++ b/pkg/metrics/backends/prometheus/prometheus.go
@@ -31,6 +31,10 @@ type Backend struct {
 	nodeLister corelistersv1.NodeLister
 }
 
+const (
+	prometheusRequestTimeout = 10 * time.Second
+)
+
 // Average CPU usage across the given nodes for the given range
 const cpuQueryTemplateString = `
 100 - (
@@ -121,8 +125,11 @@ func (b Backend) GetValue(metric string, configuration map[string]string, nodeSe
 func (b Backend) getNodeExporterPodIPsOnNodes(nodes []*corev1.Node) ([]string, error) {
 	var podIPs []string
 
-	// Filter only prom-exporter job, and further filter down by node ips
-	targets, _ := b.prometheus.Targets(context.TODO())
+	ctx, cancel := context.WithTimeout(context.Background(), prometheusRequestTimeout)
+	defer cancel()
+
+	// Filter only prom-exporter job, and further filter down by node IPs
+	targets, _ := b.prometheus.Targets(ctx)
 	for _, active := range targets.Active {
 		jobName := string(active.DiscoveredLabels["job"])
 		split := strings.Split(jobName, "/")
@@ -145,7 +152,7 @@ func (b Backend) getNodeExporterPodIPsOnNodes(nodes []*corev1.Node) ([]string, e
 func (b Backend) performQuery(query string) (float64, error) {
 	log.Debugf("Performing prometheus query: %s", query)
 
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), prometheusRequestTimeout)
 	defer cancel()
 
 	val, err := b.prometheus.Query(ctx, query, time.Time{})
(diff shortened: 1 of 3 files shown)


pkg/metrics/backends/prometheus/prometheus.go
```diff
@@ -127,25 +125,16 @@ func (b Backend) GetValue(metric string, configuration map[string]string, nodeSe
...
-	// TODO we should rethink how we do all of the below, potentially using b.Targets()
-	// to help. See https://github.com/containership/cerebral/issues/15.
-	selector := labels.NewSelector()
-	l, _ := labels.NewRequirement("prom-exporter", selection.Equals, []string{"node"})
-	selector = selector.Add(*l)
-
-	// TODO if we could use a FieldSelector including Pod.Spec.NodeName then we
-	// could avoid the gross O(n^2) below. Is there a way to do that without
-	// having to query the API directly?
-	pods, err := b.podLister.List(selector)
```
mattkelly: I believe we can remove the `podLister` from this backend entirely now, right? That would be great if possible.

pkg/metrics/backends/prometheus/prometheus.go
```diff
@@ -4,18 +4,16 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"github.com/pkg/errors"
+	"strings"
 	"text/template"
 	"time"
 
-	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
noah-vienneau: Will do thanks for letting me know

pkg/metrics/backends/prometheus/prometheus.go
```diff
@@ -127,25 +125,16 @@ func (b Backend) GetValue(metric string, configuration map[string]string, nodeSe
...
-	if err != nil {
-		return nil, errors.Wrap(err, "listing node exporter pods")
-	}
-	// Only filter in prom-exporter pods that belong to a node we care about
-	for _, pod := range pods {
-		for _, node := range nodes {
-			if pod.Spec.NodeName == node.ObjectMeta.Name {
-				podIPs = append(podIPs, pod.Status.PodIP)
+	//Filter only prom-exporter job, furhter filter down by node ips
+	targets, _ := b.prometheus.Targets(context.TODO())
```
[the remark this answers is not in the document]
noah-vienneau: Ok, I was unsure what context to use. That makes sense

pkg/metrics/backends/prometheus/prometheus.go
noah-vienneau: Looks like it is possible
Just changed in next commit

noah-vienneau: Yes removing it and a call to newClient in metrics_backend which requires a podLister argument works

mattkelly:
Thanks so much for taking this on, @noah-vienneau! Nice work. One of us from @containership/core-services will handle testing this on a real cluster. 🚀

pkg/metrics/backends/prometheus/prometheus.go
```diff
@@ -41,6 +39,9 @@ const cpuQueryTemplateString = `
 	) * 100
 )`
 
+
```
mattkelly: Remove this change

PR #117, opened 2019-06-12
metrics/backends/prometheus: Add a custom jobName field

nicholascapo:
* This allows configuration of the job name for custom Prometheus
metrics, previously the job name was hardcoded to "node-export-monitor"
* Ref #15

### DIFF
--- a/examples/autoscaling_policies/prometheus-custom-cpu-utilization.yaml
+++ b/examples/autoscaling_policies/prometheus-custom-cpu-utilization.yaml
@@ -6,8 +6,8 @@ metadata:
 spec:
   metric: custom
   metricConfiguration:
-    query: 100 - ( avg( irate(node_cpu{mode='idle',instance=~'{{.PodIPsRegex}}'}[1m]))
-      * 100)
+    query: 100 - ( avg( irate(node_cpu{mode='idle',instance=~'{{.PodIPsRegex}}'}[1m])) * 100)
+    jobName: node-exporter
   metricsBackend: prometheus
   pollInterval: 15
   samplePeriod: 300
@@ -20,5 +20,5 @@ spec:
     scaleUp:
       adjustmentType: absolute
       adjustmentValue: 2
-      comparisonOperator: '>='
+      comparisonOperator: ">="
       threshold: 70
--- a/pkg/metrics/backends/prometheus/prometheus.go
+++ b/pkg/metrics/backends/prometheus/prometheus.go
@@ -22,6 +22,10 @@ import (
 	"github.com/containership/cluster-manager/pkg/log"
 )
 
+// defaultNodeJobName is the default value used to find Prometheus Nodes
+// Specifically this is the value of the job= label that can be used to find all nodes
+const defaultNodeJobName = "node-export-monitor"
+
 // Backend implements a metrics backend for Prometheus. It requires a pod
 // lister so that it can gather info about the prom-exporter pods. Pods
 // accessed via the lister must not be mutated.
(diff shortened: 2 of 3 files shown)


pkg/metrics/backends/prometheus/prometheus.go
```diff
@@ -22,6 +22,10 @@ import (
 	"github.com/containership/cluster-manager/pkg/log"
 )
 
+// defaultNodeJobName is the default value used to find Prometheus Nodes
+// Specifically this is the value of the job= label that can be used to find all nodes
+const defaultNodeJobName = "node-export-monitor"
```
nicktate: Let's link to `https://github.com/prometheus/node_exporter` saying its the default name of the job representing the prometheus node_exporter pods.

pkg/metrics/backends/prometheus/prometheus_test.go
```diff
@@ -112,7 +112,7 @@ var (
 	dlNode1 = model.LabelSet{
 		"__meta_kubernetes_pod_node_name": model.LabelValue(promPodOnNode1.Spec.NodeName),
 		"__meta_kubernetes_pod_ip":        model.LabelValue(promPodOnNode1.Status.PodIP),
-		"job":                             "random/node-export-monitor/random",
+		"job":                             "node-export-monitor",
```
nicktate: So this test data is intended to model the results from `/api/v1/targets` from the prometheus api and I would rather keep it in a similar format vs stripping out the `/`

image

Relooking at this data, I think it actually makes sense to utilize both the `namespace` and `jobName` sections of this string.

> e.g. `containership-core/node-export-monitor/0` => `{ namespace: containership-core, jobName: node-export-monitor }`

examples/autoscaling_policies/prometheus-custom-cpu-utilization.yaml
```diff
@@ -6,8 +6,8 @@ metadata:
 spec:
   metric: custom
   metricConfiguration:
-    query: 100 - ( avg( irate(node_cpu{mode='idle',instance=~'{{.PodIPsRegex}}'}[1m]))
-      * 100)
+    query: 100 - ( avg( irate(node_cpu{mode='idle',instance=~'{{.PodIPsRegex}}'}[1m])) * 100)
+    jobName: node-exporter
```
nicktate: Let's add `jobNameNamespace: containership-core` as an example

pkg/metrics/backends/prometheus/prometheus.go
nicktate: Looking at the data from `/api/v1/targets`, I think it may make sense to add in a check for `namespace` as well to be even more flexible.

`defaultNodeJobNameNamespace` or similar

pkg/metrics/backends/prometheus/prometheus.go
```diff
@@ -131,9 +140,9 @@ func (b Backend) getNodeExporterPodIPsOnNodes(nodes []*corev1.Node) ([]string, e
 	// Filter only prom-exporter job, and further filter down by node IPs
 	targets, _ := b.prometheus.Targets(ctx)
 	for _, active := range targets.Active {
-		jobName := string(active.DiscoveredLabels["job"])
-		split := strings.Split(jobName, "/")
-		if split[1] == "node-export-monitor" {
+		job := string(active.DiscoveredLabels["job"])
+		split := strings.Split(job, "/")
+		if job == jobName || (len(split) > 1 && split[1] == jobName) {
```
nicktate: we should additionally check `split[0]` here to make sure the `namespace` is the intended namespace from the configuration

mattkelly:
@nicholascapo do you think you will have an opportunity to address the comments here soon? If not, one of us can touch it up and merge it.

nicholascapo:
Yeah, sorry, our fork of `cerebral` diverged quite a bit more than just
this.Unfortunately this means we aren't using much of this code any
more, so I don't have a way to test if it satisfies our needs.
Feel free to take it over.

mattkelly:
No worries at all @nicholascapo. Thanks again for the PR - I'll look into getting it updated and tested when I get a chance. Feel free to open issues/PRs if there's anything in your fork that you think would be useful to upstream!

nicholascapo:
Will do, thanks!

nicholascapo:
Closing since we are no longer using cerebral

